PR #31, opened 2020-08-08
adding example in python

scarrazza:
This PR implements a simple example in python.

### DIFF
--- a/examples/python-dy-aa/main.py
+++ b/examples/python-dy-aa/main.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+import pineappl
+import numpy as np
+
+
+def int_photo(s, t, u):
+    alpha0 = 1.0 / 137.03599911
+    return alpha0 * alpha0 / 2.0 / s * (t / u + u / t)
+
+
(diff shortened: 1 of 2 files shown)


scarrazza:
In principle we could create a python module with the C interface instead of distributing the `pineappl.py`.

cschwan:
- Can you complete the example so that it also, like the C example, performs the convolution?
- We could make the Python interface a little more Python-like, for example by stripping `pineappl` from every name (that's the module name already), and also strip `lumi`, `keyval`, and `grid` from the name of the functions (they should be members for the respective classes, instead of taking it as the first argument). The `pineappl_{lumi,keyval,grid}_new` functions should be constructors.
- Is it possible to make the `_delete` functions destructors that are automatically called when the objects go out of scope?

cschwan:
> In principle we could create a python module with the C interface instead of distributing the `pineappl.py`.

Sounds like a good idea.

cschwan:
This might be interesting: https://crates.io/crates/pyo3

scarrazza:
Thanks for the comments, I will open another PR with the python module implementation.
If you prefer we can merge this now.

PR #32, opened 2020-08-10
python module

scarrazza:
Here we implement the python module for PineAPPL, todo list:
- [x] installation system
- [x] use a python like API
- [x] sphinx documentation
- [x] integration with python examples
- [x] implement tests during github action

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -67,5 +67,5 @@ jobs:
     - name: Test examples
       run: |
         cd examples/python-dy-aa/
-        export PKG_CONFIG_CONFIG=/home/runner/prefix/lib/pkgconfig/:$PKG_CONFIG_PATH
+        export PKG_CONFIG_PATH=/home/runner/prefix/lib/pkgconfig/:$PKG_CONFIG_PATH
         python main.py
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -67,7 +67,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: cinstall
-        args: --prefix=$HOME/prefix --manifest-path pineappl_capi/Cargo.toml
+        args: -vv --prefix=$HOME/prefix --manifest-path pineappl_capi/Cargo.toml
     - name: Install python wrapper
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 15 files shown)


scarrazza:
@cschwan please let me know if you prefer to extend the `rules.yml` or create a new workflow for testing this module.

cschwan:
Create a new workflow `python.yaml`!

scarrazza:
Indeed, thanks. Concerning the python API, do you think we should keep or drop the keyval object allocation?

scarrazza:
Other question, do you know the prefix where the pc file is installed? 
(I just tried in `/home/runner/prefix/lib/pkgconfig` without success)

cschwan:
What would you have instead of a keyval object?

I've installed everying under `/home/runner/prefix`, see here

https://github.com/N3PDF/pineappl/blob/f8bbcf45661827cf338cc91686531eb6259da90f/.github/workflows/rust.yml#L32

scarrazza:
> What would you have instead of a keyval object?

I can create this object directly in the `grid.__init__` method, instead of passing as argument.

cschwan:
> > What would you have instead of a keyval object?
> 
> I can create this object directly in the `grid.__init__` method, instead of passing as argument.

Sure! Have a look at the following to see what values are currently supported:

https://github.com/N3PDF/pineappl/blob/f8bbcf45661827cf338cc91686531eb6259da90f/pineappl_capi/src/lib.rs#L285-L323

cschwan:
@scarrazza You can't use export, instead you have to use `echo` in the way I've shown you above.

scarrazza:
> You can't use export, instead you have to use echo in the way I've shown you above.

That's not true, the export is working fine see [here](https://github.com/N3PDF/pineappl/runs/971317026#step:12:4).

From this log the `pineappl.rc` is not stored there, I can only see the `lhapdf.pc` in `/home/runner/prefix/lib/pkgconfig`.
Do you understand why the `pineappl.rc` is not copied to the right direction?

cschwan:
Because I didn't install it ...

scarrazza:
Ok, could you please add these changes to the rust.yml in this PR?

cschwan:
Numpy is apparently missing.

cschwan:
Is anything missing or can I merge it?

scarrazza:
Yes, I think this is now completed. 
The next steps are summarized in #33 and #34 .

PR #36, opened 2020-08-12
adding workflow for pypi publication

scarrazza:
This is almost ready, however I need to wait Juan for the token.


cschwan:
Looks fine, but please always add a proper newline at the end of file (there shouldn't be a stop sign in the diff).

cschwan:
Thanks! Once we have the token we can merge it and make a point release to test it.

cschwan:
... and we should add the proper Python badges linking to the documentation and the package.

scarrazza:
Good suggestion, done.

scarrazza:
@cschwan this is ready now, however before merging let me check the convolution in python, I will let you know.

Here the projects:
- https://pypi.org/project/pineappl/
- https://pineappl.readthedocs.io

However, for pypi I had to upload manually the first release (0.2.0) in order to create the project and define a token.

scarrazza:
@cschwan, I would like to merge #38 here (it is related to conda).

However, I noticed that we do not provide any example/tests about how to load the generated '*.pineappl' grid and perform the convolution. I can't find the capi call for this feature, but I believe would be great to have it here.

cschwan:
@scarrazza It's just `pineappl_grid_read`, see [here](https://docs.rs/pineappl_capi/0.2.0/pineappl_capi/fn.pineappl_grid_read.html).

scarrazza:
So this function fills the grid, is there some way to recover the number of bins and keyval? I believe we need at least the number of bins in order to perform the convolution.

cschwan:
Yes, the number of bins you can get by calling [`pineappl_grid_bin_count`](https://docs.rs/pineappl_capi/0.2.0/pineappl_capi/fn.pineappl_grid_bin_count.html). Getting the bin limits doesn't seem to be supported by the C API, but you can get the bin sizes with [`pineappl_grid_sizes`](https://docs.rs/pineappl_capi/0.2.0/pineappl_capi/fn.pineappl_grid_bin_sizes.html). Reading the key-value pairs you can by calling [`pineappl_keyval_bool`](https://docs.rs/pineappl_capi/0.2.0/pineappl_capi/fn.pineappl_keyval_bool.html), [`pineappl_keyval_double`](https://docs.rs/pineappl_capi/0.2.0/pineappl_capi/fn.pineappl_keyval_double.html), and so on, but there's function to get the keyval object from the grid.

cschwan:
Let me know which functions you need and I'll implement them.

scarrazza:
Ok, I think we can just have the bins, I just have to check if there is some way to convert the `uintptr_t` to `int` via ctypes.

scarrazza:
This is ready. At this point I need a new release 0.2.1 in order to implement conda.
Please remember that the python wrapper hardcodes the version in `wrappers/python/src/pineappl/__init__.py`.

PR #38, opened 2020-08-13
implementing pytests

scarrazza:
This completes the python wrapper with pylint and pytest.


cschwan:
Does the test allow for small variations around the numerical values? See for instance the rust test case, https://github.com/N3PDF/pineappl/blob/1e9b13e248318de3943f5bcb6674a34d87e5cc7d/pineappl/tests/drell_yan_lo.rs#L204-L206 where I allow for a variation of 16 ULPS. If you just comparing the standard output, you could simply print and compare less digits.

scarrazza:
I am using the numpy assert all close with rtol=1e-5 [here](https://github.com/N3PDF/pineappl/pull/38/files#diff-c25c29d9c363b1834203695062b35f51R30), I believe this should be enough.

cschwan:
I see; in that case you can make the test even stricter, something around 1e-12?

scarrazza:
This should be ready now.

cschwan:
OK, merge it!

PR #72, opened 2021-09-14
Release Python Workflow

AleCandido:
It might work as it is, but:

- [x] if possible switch from `${{ secrets.PYPI_PWD }}` (yet to be provided) to `${{ secrets.PYPI_TOKEN }}` (already present)
- [x] test on [PyPI test](https://test.pypi.org/)
  - even locally would be fine
  - but this is not possible for macOS (of course...), then find a way to use `maturin` with [PyPI test](https://test.pypi.org/) and temporarily replace event with [`workflow_dispatch`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch)
- [x] decide a name: the name of the PyPI package will be the one of the crate specified in `pineappl_py/Cargo.toml` (thus `pineappl` it's not available, even though it will be the name of the python package used with `import`, since it is the one used in `pineappl_py/src/lib.rs`)
https://github.com/N3PDF/pineappl/blob/65e3c137cd9aba1cc4c5a3c84312820328c58582/pineappl_py/Cargo.toml#L2
https://github.com/N3PDF/pineappl/blob/65e3c137cd9aba1cc4c5a3c84312820328c58582/pineappl_py/src/lib.rs#L12-L13


AleCandido:
The linux release has been tested manually using the container, but both workflows are not tested because it will only be possible after merging with `master`.

cschwan:
Concerning the name I've found this: https://github.com/PyO3/maturin/issues/313#issuecomment-784438527 - can you try that?

AleCandido:
I will do as soon as possible

AleCandido:
Now it is actually working, producing wheels for the `pineappl` python package :D

Issue #73, opened 2021-10-01
Allow to easily generate order masks


scarlehoff:
In order to mask orders in the python interface I need to pass five different booleans

```python
    qcd_result = pine.convolute(xf, xf, a_s,  [True, True, False, False, False])
```

What is the combination for NLO QCD and NLO EW? (so, LO * as and LO * a)
The closest combination for the he qcd_result I get is `[True, False, False, False, True]` but would prefer to have the exact definition.

(wasn't sure whether this is a python or internal pineapl thing, so wasn't sure whether to direct the question to @cschwan or @felixhekhorn @AleCandido)

cschwan:
You need to match the ordering returned by `grid.orders()`. This returns a list of [`PyOrder`](https://pineappl.readthedocs.io/en/latest/modules/pineappl/pineappl.html#pineappl.pineappl.PyOrder) objects, which you can query with `as_tuple()`, which in turn returns the exponents. For Drell-Yan you should get the following tuples:

    (0, 2, 0, 0)
    (1, 2, 0, 0)
    (1, 2, 1, 0)
    (1, 2, 0, 1)
    (0, 3, 0, 0)
    (0, 3, 1, 0)
    (0, 3, 0, 1)

The ordering can be different, and some tuples may be missing. The first tuple is the `alpha^2` contribution (that's the LO), the next three are the NLO QCD (one power in `alphas^1`), and the last three are the NLO EW (`alpha^3`). If you don't need scale variations you can ignore (mask out) all orders that have non-zero power in `logxir` or `logxif`.

To calculate NLO QCD, you would therefore have to construct the following mask (no scale variations):

    [True, True, False, False, False, False, False]

NLO EW would be

    [True, False, False, False, True, False, False]

Note again that the ordering of the tuples given above are arbitrary, and that there can be many NLO contributions depending on the process, jets for instance can have four different NLO (and three LO), each with a renormalization and factorization scale variation grid, yielding in total 12 NLO grids (and three LO grids).

scarlehoff:
I guess for different DY I would get different orders as well? For instance, for one of the DY_13TeV I get:

```
[orde.as_tuple() for orde in pine.orders()]

[(0, 2, 0, 0),
 (0, 3, 0, 0),
 (0, 3, 1, 0),
 (0, 3, 0, 1),
 (1, 2, 0, 0),
 (1, 2, 1, 0),
 (1, 2, 0, 1)]
```

[...]

Issue #66, opened 2021-05-06
Fix `Grid::convolute` caching strategy


cschwan:
Right now `Grid::convolute` always has `two_caches = true` (I've discovered this while running the code in `rust-gdb`),

https://github.com/N3PDF/pineappl/blob/6ccb9a49fa9d3baa06f9958fd1db6f9286983449/pineappl/src/grid.rs#L343

which misses a few optimizations whenever the two initial-state hadron are the same.

cschwan:
Plan:

- [ ] replace arguments `xfx1`, `xfx2` and `alphas` with a single object of `PdfCache`,
- [ ] which takes care of caching the arguments of every PDF,
- [ ] in such a way that pp, ppbar, etc. rely on a single PDF.

cschwan:
A similar benchmark with `pdf_uncertainty` instead of `convolute` takes 1m37s in the first case and 4m9s in the second case; this subcommand uses 8 cores. This just goes to show that PDF caching is still very much needed.

PR #74, opened 2021-10-02
Python interface improvements

AleCandido:
While working on stuffs that make use of python interface, new needs arose. Thus, there is room to add some improvements.

Following a list of possible ones:
- [ ] a `Grid.get()` method (or even better that one, plus a `Grid.__getitem__()` method, and then also `Grid.__setitem__` to alias `Grid.set_key_value()`), see https://github.com/NNPDF/runcards/pull/104#issuecomment-932754107
- [ ] a function close to `merge` subcommand see https://github.com/NNPDF/runcards/pull/104#issuecomment-932754107
  - we can think if we want to offer a module to programmatically use the CLI
- [ ] maybe a function to generate order masks from something similar to "NLO QCD" or "NLO EW", see #73


AleCandido:
At the moment the base for the PR is fake, because there is no commit. As soon as I commit something I'll switch to master.

AleCandido:
As discussed with @felixhekhorn we'll reserve `Grid.__get_item__()` and `Grid.__set_item__()` and `Grid.__iter__()` for managing subgrids, since `Grid` is a subgrids container in the first place (and a metadata container only afterwards).

AleCandido:
Most of the functions of CLI consists mainly in printing tables, something we don't need for the python interface (since we can handle objects, and it is comfortable to do it), nevertheless we are wondering if it's worth to bind `pineappl_cli.helpers`.

cschwan:
> Most of the functions of CLI consists mainly in printing tables, something we don't need for the python interface (since we can handle objects, and it is comfortable to do it), nevertheless we are wondering if it's worth to bind `pineappl_cli.helpers`.

If you find that helpful I'd rather suggest to import the functionality into PineAPPL and remove the corresponding helper (and write a wrapper to the new function in PineAPPL). I'm definitely going to do that for the `convolute` function, which takes care of charge conjugation and different initial states; this also overlaps with #66. Let me know if you need more.

AleCandido:
> > Most of the functions of CLI consists mainly in printing tables, something we don't need for the python interface (since we can handle objects, and it is comfortable to do it), nevertheless we are wondering if it's worth to bind `pineappl_cli.helpers`.
> 
> If you find that helpful I'd rather suggest to import the functionality into PineAPPL and remove the corresponding helper (and write a wrapper to the new function in PineAPPL). I'm definitely going to do that for the `convolute` function, which takes care of charge conjugation and different initial states; this also overlaps with #66. Let me know if you need more.

I would say just `convolute()` (and consequently `SCALES_VECTOR`). I can do here or we can do separately, that's fine.

You can move yourself if you wish, or I can move it, but where should I place? and how should I call it? (my guess would be to put it in `grid.rs` as a separate functions, i.e. not inside the `impl` block, but I don't have any good name other than `convolute`).

cschwan:
> > > Most of the functions of CLI consists mainly in printing tables, something we don't need for the python interface (since we can handle objects, and it is comfortable to do it), nevertheless we are wondering if it's worth to bind `pineappl_cli.helpers`.
> > 
> > 
> > If you find that helpful I'd rather suggest to import the functionality into PineAPPL and remove the corresponding helper (and write a wrapper to the new function in PineAPPL). I'm definitely going to do that for the `convolute` function, which takes care of charge conjugation and different initial states; this also overlaps with #66. Let me know if you need more.
> 
> I would say just `convolute()` (and consequently `SCALES_VECTOR`). I can do here or we can do separately, that's fine.
> 
> You can move yourself if you wish, or I can move it, but: where should I place? and how should I call it? (my guess would be to put it in `grid.rs` as a separate functions, i.e. not inside the `impl` block, but I don't have any good name other than `convolute`).

My current idea is that instead of passing three functions (`xfx1`, `xfx2` and `alphas`) you pass a single `PdfCache` object (for which I have yet to write the struct). You can then easily cache PDFs, even among several `convolute` calls, which will speed up a few of the subcommands.

AleCandido:
Added syntactic sugar to get and set subgrids, but the `get` function is not really working well.

The following code:
```py
import pineappl
g = pineappl.grid.Grid.read("dyaa.pineappl")
s = g[0, 0, 0]
g[0, 1, 0] = s
```
(with `dyaa.pineappl` the grid generated from [the example](https://github.com/N3PDF/pineappl/blob/py-improved/examples/python/dyaa.py)) raises with the following error
```
AttributeError                            Traceback (most recent call last)
<ipython-input-5-346819bef8ca> in <module>
----> 1 g[0, 1, 0] = s

~/Projects/N3PDF/pineappl/pineappl_py/env/lib/python3.9/site-packages/pineappl/grid.py in __setitem__(self, key, subgrid)
    140             raise ValueError("A tuple with `(order, bin, lumi)` is required as key.")
    141
--> 142         self.set_subgrid(*key, subgrid)
    143
    144     def set_remapper(self, remapper):

~/Projects/N3PDF/pineappl/pineappl_py/env/lib/python3.9/site-packages/pineappl/grid.py in set_subgrid(self, order, bin_, lumi, subgrid)
    122                 subgrid content
    123         """
--> 124         self.raw.set_subgrid(order, bin_, lumi, subgrid.into())
    125
    126     def __setitem__(self, key, subgrid):

AttributeError: 'builtins.PySubgridEnum' object has no attribute 'into'
```

Essentially the error is due to the subgrid not having been recast to any specific subgrid type when retrieved, so the `subgrid.into()` call is unneeded. I believe it should be recast to a meaningful type to be able to do operations, but since I don't know which one should be maybe that's a hint I'm not thinking to the correct solution.

cschwan:
> As discussed with @felixhekhorn we'll reserve `Grid.__get_item__()` and `Grid.__set_item__()` and `Grid.__iter__()` for managing subgrids, since `Grid` is a subgrids container in the first place (and a metadata container only afterwards).

Do `__get_item__` and others significantly improve the usability of `Grid`? I'm a bit sceptical here; I'd actually like the user to avoid using subgrids at all, and everything should be done using `convolute`, `convolute_subgrid` and `convolute_eko`. This doesn't always work (for instance in `yadism`), but that should be the aim.

AleCandido:
> Do `__get_item__` and others significantly improve the usability of `Grid`? I'm a bit sceptical here; I'd actually like to discourage the use of subgrids, because everything should be done using `convolute`, `convolute_subgrid` and `convolute_eko`. This doesn't always work (for instance in `yadism`), but that should be the aim.

Ok, this might be a point. Nevertheless, I wonder if giving write access without read access is a good idea (about `__setitem__` instead is just for building abstractions, in python it is useful to build on existing abstractions, and if a `Grid` is a subgrid containers you expect to be able to access them, set them and iterate over them, doing it with standard functions helps the user to rely on known abstractions).

In case I can just revert and remove `Grid.subgrid()`.

AleCandido:
Instead I'm running in more serious troubles with `Grid.merge()`.

We decided to keep `Grid` non-clonable, and I agree it is a good idea. But for `Grid.merge()` we need to pass a second grid, and it looks like it doesn't like:
```
error[E0277]: the trait bound `PyGrid: Clone` is not satisfied
   --> src/grid.rs:345:40
    |
345 |     pub fn merge(&mut self, mut other: Self) {
    |                                        ^^^^ the trait `Clone` is not implemented for `PyGrid`
    |
    = note: required because of the requirements on the impl of `pyo3::FromPyObject<'_>` for `PyGrid`
```

Any suggestion?

cschwan:
> Instead I'm running in more serious troubles with `Grid.merge()`.
> 
> We decided to keep `Grid` non-clonable, and I agree it is a good idea. But for `Grid.merge()` we need to pass a second grid, and it looks like it doesn't like:
> 
> ```
> error[E0277]: the trait bound `PyGrid: Clone` is not satisfied
>    --> src/grid.rs:345:40
>     |
> 345 |     pub fn merge(&mut self, mut other: Self) {
>     |                                        ^^^^ the trait `Clone` is not implemented for `PyGrid`
>     |
>     = note: required because of the requirements on the impl of `pyo3::FromPyObject<'_>` for `PyGrid`
> ```
> 
> Any suggestion?
> 
> P.S.: `PyGrid` is not clonable because it contains a `Grid`, that as said above it's not clonable

That's a tough one, let me think about it.

felixhekhorn:
Mmm, I think the only solution (if we want to keep `Grid` non-clonable) is to provide a function `merge_from_file(Grid, path)` and then read the `Grid` in Rust ...

AleCandido:
I'm on my way to remove the `additional_metadata`, but since I had to update `EkoInfo` (that now has a different meaning), I took the chance to reshape it in a more consistent way.

Can you review and make your comments? @cschwan

cschwan:
> Mmm, I think the only solution (if we want to keep `Grid` non-clonable) is to provide a function `merge_from_file(Grid, path)` and then read the `Grid` in Rust ...

Yes, I agree this is how we should do it!

cschwan:
The item concerning the convolute helpers now basically amounts to writing a Python wrapper to `LumiCache`. I need to replace every call to `convolute` with a corresponding call to `convolute2` (and after this is done rename `convolute2` to `convolute`), which will require a `LumiCache`. This cache you can create with one PDF set (`LumiCache::with_one`) or with two PDF sets (`LumiCache::with_two`). You'll need only one PDF set for all DIS processes, for most LHC processes, and for Tevatron and similar colliders, because the anti-proton PDF is automatically generated through charge conjugation in `LumiCache`. Two PDF sets are really only needed for processes like proton-lead collisions.

I'll try to make the necessary adjustments to the Python wrapper myself.

PR #75, opened 2021-10-05
Support evolution basis FkTables

AleCandido:
We want to support the generation of FkTables in evolution basis, in order to get closer to the old FkTables.

Since the rotation is already available inside `eko`, the only thing missing was to allow for different input and output basis during convolution of pineappl grids.
(actually the only information relevant for the inner convolution is the length of the basis, that has to be the same for input and output, so the only thing left is to propagate the information on output basis to match it correctly with the PDF later on)

### DIFF
--- a/pineappl/src/grid.rs
+++ b/pineappl/src/grid.rs
@@ -1141,6 +1141,7 @@ impl Grid {
         alphas: &[f64],
         (xir, xif): (f64, f64),
         pids: &[i32],
+        target_pids: &[i32],
         x_grid: Vec<f64>,
         q2_grid: Vec<f64>,
         operator: Array5<f64>,
@@ -1178,7 +1179,6 @@ impl Grid {
             _ => unimplemented!(),
         };
 
-        // create target luminosities
         let pids1 = if has_pdf1 {
             pids.to_vec()
         } else {
@@ -1189,9 +1189,20 @@ impl Grid {
         } else {
             vec![initial_state_2]
         };
-        let lumi: Vec<_> = pids1
+        // create target luminosities
+        let tgt_pids1 = if has_pdf1 {
+            target_pids.to_vec()
+        } else {
+            vec![initial_state_1]
+        };
+        let tgt_pids2 = if has_pdf2 {
+            target_pids.to_vec()
+        } else {
+            vec![initial_state_2]
+        };
+        let lumi: Vec<_> = tgt_pids1
             .iter()
-            .cartesian_product(pids2.iter())
+            .cartesian_product(tgt_pids2.iter())
             .map(|(a, b)| lumi_entry![*a, *b, 1.0])
             .collect();
 


cschwan:
@felixhekhorn + commit c42860c: Please don't do that, instead add metadata to the `Grid` that `convolute_eko` creates. The argument list of `convolute_eko` is already far too long.

felixhekhorn:
> @felixhekhorn + commit [c42860c](https://github.com/N3PDF/pineappl/commit/c42860c5477b3564cf0202f5f20ea0233afc35ef): Please don't do that, instead add metadata to the `Grid` that `convolute_eko` creates. The argument list of `convolute_eko` is already far too long.

We thought about it, but we do like the alternative even less:

- to my understanding FkTables should be immutable, i.e. read-only, because you're never supposed to change them once they are created (e.g. `set_key_value` should not be allowed)
- and @AleCandido correctly said that we would also have a simply wrong FkTable in the middle ...

AleCandido:
> @felixhekhorn So what we want is to add _very specific_ metadata to the `FkTable`, which when it is missing or wrong is an error. In that case we better add new parameters to the arguments of `convolute_eko`. The advantage of that over a `HashMap` is that the metadata has to be given explicitly and is strongly typed.
> 
> We should probably pack all the arguments, except maybe the operator itself into a new `struct` to keep the number of the arguments small (that's independent of of the discussion above, but adding more arguments compounds the problem).

I agree with both.

Later on we should also improve a bit the `convolute_eko` function itself, since at the moment it's barely readable. I know it's a complex operation, but maybe we can refactor with some tinier functions to perform more atomic tasks (as I said, later on...).

cschwan:
@AleCandido thanks, the `EkoInfo` struct was really wrongly named, but now it correctly reflects the fact that it's information from EKO.

AleCandido:
Yes, on the other hand I'm not 100% convinced about `GridAxes`, because at some point (hope never) we might need to pass EKO more information, that is not only the axes.

Nevertheless, we can say that for the time being is fine, and maybe it'll be fine forever.

AleCandido:
Added a very minimal (and working) test in be69003, but at least it guarantees that `convolute_eko` keeps running.

AleCandido:
At this point:
- we replaced the `convolute_eko` input (in rust) from a lot oaf arguments to a single struct + an array
- in python still there is a plethora of arguments for `convolute_eko`, but I'm not sure that we want to expose `EkoInfo` also there, and the actual public python API (the exposed to the user), actually takes an `eko.Output` object as an input, so everything is inferred (`eko.Output` goes into many arguments, those are unpacked to `EkoInfo` + array)
- we have improved test coverage

Now I'm in favor to merge.

AleCandido:
We should optimize evolution basis products when they are zero by cancellation (i.e. zero more or less at float precision)

Issue #89, opened 2021-11-17
Improve the CLI's documentation of the `remap` subcommand


felixhekhorn:
To continue the APPL comparison I will need to renormalize the grids - I saw that the `remap` command seems to be able to do such a thing, but I don't know how to use it; i.e. what is the "remap string" - can you provide us with an example?

cschwan:
The primary purpose of `remap` is to map a one-dimensional distribution to an n-dimensional one (many Monte Carlo programs only produce one-dimensional distributions, and `remap` is our way to work around it). I'm not quite sure this is the proper use in this case, but let's try it anyway to see what you need.

You can find an example here:

https://github.com/NNPDF/runcards/blob/a48e5d4c7c6d2dae36f0a0228a11e4739f87f7be/nnpdf31_proc/ATLAS_DY_8TEV_2D_0116_0200/postrun.sh#L3

What it does is to completely ignore whatever bin limits you previously set, and apply new ones, based on a cartesian product. In the case above, '116,150,200,...;0,0.2,0.4,...' produces the two-dimensional bin limits in the following order:

- from 116 to 150 for the first observable, and from 0.0 to 0.2 for the second observable,
- 116-150 and 0.2-0.4,
- ...
- 150-200 and 0.0-0.2,

The pipe `|` is used to modify the limits for the second observable for specific values of the first observable. If you count the number of fields divided by the pipe you'll find five, one field for each of the five bins of the first observable. An empty field just means the field previously specified. In the example above that means specifically

- 300-500 and 0.0-0.4,
- 300-500 and 0.4-0.8,
- ...
- 500-1500 and 0.0-0.4,
- 500-1500 and 0.4-0.8,
- ...

By default `remap` divides the result of each bin by the bin-width (for each observable). If you don't want that, you can disable it for specific observables using `--ignore-obs-norm index` where `index` is the index of the observable starting from `0`. An example for that you can find here,

https://github.com/NNPDF/runcards/blob/master/nnpdf31_proc/ATLAS_DY_8TEV_3D_046080_0007/postrun.sh

where only the bin width for the first (`0`) observable is taken into account.

Finally, you can use `--norm factor`, which applies `factor` multiplicatively for all bins (for example for the femtobarn/picobarn conversion).

AleCandido:
Maybe @cschwan it might be an idea to extend CLI help with examples, at some point.

[...]

PR #90, opened 2021-11-17
Apply grid normalization

AleCandido:
Hi @cschwan, as you know from #89 we are dealing with normalizations, and I proposed to skip `pineappl remap` and write instead something in python.

The idea is:
1. load the grid
2. apply normalization
3. dump the grid

Nevertheless, point 2 is not that easy, so I tried to figure a way out of it, as easy as possible.

I had a look inside `pineappl remap`, and I found that you use `BinRemapper` to apply normalization:
https://github.com/N3PDF/pineappl/blob/32e63c92cd9ea8358d6e193e068a9ee33d917979/pineappl_cli/src/remap.rs#L155-L161
and this is the only place where the CLI argument `norm` appears.

So my idea for point 2 would be:
1. get the remapper (`grid.more_members.upgrade()` should guarantee is available)
2. copy the limits
3. copy normalizations and multiply by norm 
4. create a new `BinRemapper`
5. set the new remapper

Unfortunately, I noticed the `remapper` inside the `more_members` is an option, so it might not be available, then: what should I do in the case?

Moreover, I would need to implement a getter for `bin_remapper` in python, and before even in `pineappl` crate. Then I was wondering:
- should I implement the getter and do as above?
- or should we make the `normalize`/`renormalize` a method of `Grid` directly inside `pineappl`?

P.S.: I needed a commit to open the PR, but I had nothing to commit, so the moment we start I'll just squash it and remove the empty `help` file

### DIFF
--- a/pineappl_py/src/grid.rs
+++ b/pineappl_py/src/grid.rs
@@ -465,4 +465,14 @@ impl PyGrid {
             })
             .collect()
     }
+
+    /// Scale all subgrids.
+    ///
+    /// Parameters
+    /// ----------
+    /// factor : float
+    ///     scalar factor by which scaling
+    pub fn scale(&mut self, factor: f64) {
+        self.grid.scale(factor);
+    }
 }


cschwan:
The situation with `BinRemapper` is unfortunately complicated, as I didn't forsee something like it would be needed when I designed the `Grid` struct. I think what you want to do is to use `BinInfo`, for which you can get an instance using `Grid::bin_info`. With a `BinInfo` object you can get all limits using `BinInfo::left` and `BinInfo::right`, and also the normalizations using `BinInfo::normalizations`, irrespectively of whether there's a remapper or not. This you can use to a set a remapper.

cschwan:
See also https://github.com/N3PDF/pineappl/issues/83#issue-1039614822.

AleCandido:
I took the chance to add `grid.scale` to python API.

It's doing nothing else, then I would merge.

cschwan:
All right, go ahead!

PR #95, opened 2021-12-06
Improve PineAPPL's CLI

cschwan:
This branch is supposed to rework the CLI to

- [x] use `clap`'s derive feature that makes it much easier to write an easy-to-understand subcommand and
- [ ] to make sure no regressions are introduced by testing each of the subcommands;
  - [ ] `channels`
  - [ ] `convolute`
  - [ ] `diff`
  - [ ] `info`
  - [ ] `luminosity`
  - [ ] `merge`
  - [ ] `optimize`
  - [ ] `orders`
  - [ ] `pdf_uncertainty`
  - [ ] `plot`
  - [ ] `pull`
  - [ ] `remap`
  - [ ] `set`
  - [ ] `subgrids`
  - [ ] `sum`
  - [ ] `upgrade`


AleCandido:
Are you really sure you want to test the full hardcoded docstring? If at some point you'll want to update `clap` version and they even change something minor (say spacing) it will break all the tests...

cschwan:
The `--help` tests are just a basic implementation of the more general idea to test coverage of all subcommands. Note that the version of `clap` is pinned.

AleCandido:
I agree that `clap` won't update on its own (and that's a good thing), nevertheless at some point you might want to do it manually (and this was the meaning of the original observation).

I agree that coverage is a good thing, but nor 100% is granting to be bug free, neither any other reason exists to make 100% strictly required: for as long as you know what you are testing, and more important what not, that's perfectly fine (but this of course is only my personal point of view).

cschwan:
My point is more that we don't have any tests for the CLI, which is potentially dangerous. While writing tests I already discovered a problem that is also present in master (fixed in commit 10c659ea6631b87a0bd86e0a76b9f934baf5de23). I agree that the `--help` tests are unnecessarily fragile, but once we have proper tests we can probably remove them.

PR #103, opened 2022-01-19
Speedup `Grid::convolute_eko`

cschwan:
The runtime for `Grid::convolute_eko` for double-hadronic grids is, unfortunately, quite long, of the order of 10-15 minutes for a typical case. I'm convinced that we can improve the runtime, and I have at least two ideas how:

1. Eliminate zeros of the EKO early. Commit 6aa3ba4668df63d22b41b0c82ee59e4e65df5280 checks, for all source- and target-PID tuples, whether the operator is non-zero by going through all its elements once. For instance, in many of the grids we have a photon as an initial state, but often we ignore the PDF at the fitting scale and so we can skip everything that has a photon. This is also true for top and anti-top quarks, and possibly for other quark flavour as well.
2. The actual convolution can be rewritten as a loop over the different mu2 values, each performing the following operation in the general case where there are two hadronic initial states: `(FK)_ij = sum_k O_il (G_lm)_k O_mj`, so that we could leverage fast linear algebra libraries.

However, we need to test and benchmark this.

### DIFF
--- a/pineappl/src/grid.rs
+++ b/pineappl/src/grid.rs
@@ -1608,12 +1608,14 @@ impl Grid {
 
                         // if `op1` and `op2` below are zero there's no work to do
                         // TODO: ideally we change the for loops instead of vetoing here
-                        if (has_pdf1 && !non_zero_pid_indices
-                            .iter()
-                            .any(|&tuple| tuple == (tgt_pid1_idx, src_pid1_idx)))
-                            || (has_pdf2 && !non_zero_pid_indices
+                        if (has_pdf1
+                            && !non_zero_pid_indices
                                 .iter()
-                                .any(|&tuple| tuple == (tgt_pid2_idx, src_pid2_idx)))
+                                .any(|&tuple| tuple == (tgt_pid1_idx, src_pid1_idx)))
+                            || (has_pdf2
+                                && !non_zero_pid_indices
+                                    .iter()
+                                    .any(|&tuple| tuple == (tgt_pid2_idx, src_pid2_idx)))
                         {
                             continue;
                         }
--- a/pineappl/src/grid.rs
+++ b/pineappl/src/grid.rs
@@ -1608,10 +1608,10 @@ impl Grid {
 
                         // if `op1` and `op2` below are zero there's no work to do
                         // TODO: ideally we change the for loops instead of vetoing here
-                        if non_zero_pid_indices
+                        if !non_zero_pid_indices
                             .iter()
                             .any(|&tuple| tuple == (tgt_pid1_idx, src_pid1_idx))
-                            || non_zero_pid_indices
+                            || !non_zero_pid_indices
                                 .iter()
                                 .any(|&tuple| tuple == (tgt_pid2_idx, src_pid2_idx))
                         {
--- a/pineappl/src/grid.rs
+++ b/pineappl/src/grid.rs
@@ -1440,6 +1440,39 @@ impl Grid {
             })
             .collect();
 
+        // the x1 and x2 grid values are the same across all non-zero subgrids
+        let first_non_zero_grid = self
+            .subgrids
+            .iter()
+            .filter(|subgrid| !subgrid.is_empty())
+            .next()
+            .unwrap_or_else(|| unreachable!());
+        // source x1/x2 grid might differ and be differently sorted than the operator
+        let x1_grid: Vec<_> = first_non_zero_grid
+            .x1_grid()
+            .iter()
+            .map(|x| {
+                eko_info
+                    .grid_axes
+                    .x_grid
+                    .iter()
+                    .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
+                    .unwrap_or_else(|| unreachable!())
+            })
+            .collect();
+        let x2_grid: Vec<_> = first_non_zero_grid
+            .x2_grid()
+            .iter()
+            .map(|x| {
+                eko_info
+                    .grid_axes
+                    .x_grid
+                    .iter()
+                    .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
+                    .unwrap_or_else(|| unreachable!())
+            })
+            .collect();
+
         // iterate over all bins, which are mapped one-to-one from the target to the source grid
         for bin in 0..self.bin_info().bins() {
             // iterate over the source grid luminosities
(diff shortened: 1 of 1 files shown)


felixhekhorn:
@cschwan I added an additional condition to the skipping to make it DIS compatible

cschwan:
@felixhekhorn cheers, good catch!

felixhekhorn:
On `ATLAS_TTB_8TEV_TOT` we went from 3:15 to 2:51, so not that much ... fine for me to merge

felixhekhorn (approved):

AleCandido (approved):

cschwan:
@AleCandido @felixhekhorn that was just 1), but the my hope is that 2) gives a much larger improvement (not yet implemented).

cschwan:
I observed that the EKO for the default grid in Pineko is triangular in the source and target `x`, but the row with source `x` has 3 more entries than expected for a truly triangular matrix. That's probably because of the interpolation order (3?). However, this is different for points close to `x = 1` (5 more entries). Did you increase the interpolation order in this region?

felixhekhorn:
As explained in the [docs](https://eko.readthedocs.io/en/latest/theory/Interpolation.html#example) we're taking the N_deg+1 (default: 4+1=5) nearest points - which on the border leads to *additional* offdiagonals

felixhekhorn:
>     2. The actual convolution can be rewritten as a loop over the different mu2 values, each performing the following operation in the general case where there are two hadronic initial states: `(FK)_ij = sum_k O_il (G_lm)_k O_mj`, so that we could leverage fast linear algebra libraries.
> 
> 
> However, we need to test and benchmark this.

Can we postpone 2. to a separate PR and close this one?

AleCandido:
> Can we postpone 2. to a separate PR and close this one?

Maybe convert 2. in an issue first, since no one has time to dedicate to it right now.

cschwan:
Or we simply leave it open because now this PR basically is 2 :smile:. Why do you want to close it?

AleCandido:
> Or we simply leave it open because now this PR basically is 2 smile. Why do you want to close it?

Fine by me.
I believe @felixhekhorn wanted to start using 1., but if I remember correctly it wasn't a big change in performance.

felixhekhorn:
this is getting stale (no activity the last half year) and the longer you wait the more difficult it will be merge

pineappl/src/grid.rs
```diff
@@ -1595,6 +1613,20 @@ impl Grid {
...
+                        if (has_pdf1
+                            && !non_zero_pid_indices
+                                .iter()
+                                .any(|&tuple| tuple == (tgt_pid1_idx, src_pid1_idx)))
+                            || (has_pdf2
+                                && !non_zero_pid_indices
+                                    .iter()
+                                    .any(|&tuple| tuple == (tgt_pid2_idx, src_pid2_idx)))
+                        {
+                            continue;
```
felixhekhorn, line 1627: @cschwan if we assume for a second the error is in here, this `continue` is realistically the only option for something going wrong. So,
- the comment says "if `op1` **and** `op2`", but the condition features an **or** between two things - does this make sense?
- the other option is `non_zero_pid_indices` being computed wrongly or used wrongly, which could explain why some grids are working and others not

I don't believe this to be a problem, but "the x1 and x2 grid values are the same across all non-zero subgrids" - is this statement ever asserted?

cc @andreab1997

felixhekhorn: > * the comment says "if `op1` **and** `op2`", but the condition features an **or** between two things - does this make sense?

Actually, I guess the comment is wrong: it should be "if either `op1` or `op2` is zero there's no work to do"

felixhekhorn: > the other option is non_zero_pid_indices being computed wrongly

now that I'm thinking about it: @andreab1997 another possible source of a bug might be the flavor rotation inside pineko: because we rotate there the eko from flavor-flavor to evolution-flavor as FK tables are by default in evolution space, but eko and grids are by default in flavor space only (actually, eko is a tensor so it has to set of indices). But again since it is working for some datasets this is unlikely. However, this would a possibility of eko being correct, yet FK wrong.

andreab1997: > now that I'm thinking about it: @andreab1997 another possible source of a bug might be the flavor rotation inside pineko: because we rotate there the eko from flavor-flavor to evolution-flavor as FK tables are by default in evolution space, but eko and grids are by default in flavor space only (actually, eko is a tensor so it has to set of indices). But again since it is working for some datasets this is unlikely. However, this would a possibility of eko being correct, yet FK wrong.

Yes indeed is unlikely but I can try to test it.

cschwan: @felixhekhorn the next commit adds negations, and I think it should be an OR instead of an AND, at least for the double hadronic case. Can you confirm or deny this!

felixhekhorn: I think that condition might be fine as is - as said above it's only the comment which is wrong - moreover, we just tried to just remove the `continue` and it seems nothing changes, i.e. this might not be the offending change - unfortunately ... (actually, even in practice the condition should be almost always `false`). However, now we also suspect that even DIS might be affected (@andreab1997 can maybe confirm) which is easier to check.

andreab1997: Actually I just tested NMCPD_dw_ite (only NMC_NC_EM_P_F2.pineappl.lz4) and it worked. I am going to try also the other DIS fks but at the moment I would say that the bug does not affect DIS.

andreab1997: I have tried about 10 different DIS FKs and they are all correct with pineko==0.2.2 and pineappl==0.5.5

cschwan: I'm now convinced that commit b7014fdc4ecf9ed3ef15f08cc47ee5761ed0b471 introduces a bug in the case where the x-grids of processes change over bins, orders or luminosities. This usually happens in the case we convert them from old APPLgrids. @andreab1997 do you a need new release to test a different version of PineAPPL or can you also test it from git?

AleCandido: Since it has been modified a bit, I'd propose to revert it manually in a new branch, and test if that's the issue.

andreab1997: I can test it from git but today I am a bit busy with seminars. If you can test it thank you, otherwise I will test it soon as possible.

AleCandido: I will do the commit, and @felixhekhorn has a quick test already in place

cschwan: Alright, thanks to everyone for figuring this out! I will mark this discussion as resolved, and everything else we'll discuss in #182.

PR #105, opened 2022-01-28
Support native pathlike objects

### LINKED ISSUE #84: Python paths handling
I don't remember any longer where we wrote it, but it was already pointed out by @cschwan (I guess, otherwise @felixhekhorn) that since July it is possible to [handle `pathlib.Path` directly from `pyo3`](https://pyo3.rs/v0.14.5/changelog.html?highlight=path#added-5).

Thus, we should rework the API when dealing with paths, to work with python (and rust) paths, instead of relying on strings.

In particular, it is relevant for `Grid::{read,write}` and `FkTable::{read,write}`.

AleCandido:
This should close #84.

It's a tiny PR, and we can even quickly merge.
I ran existing tests, and slightly extended (sorry for the diff, but someone forgot to run `black` on unit tests).
I even tested it manually, and it's smoothly working.

Speaking of tests: maybe it might be a good chance to extend the IO related ones a bit (`write_lz4` is not tested, nor `read`, `write` and `write_lz4` for FkTables... actually almost nothing is tested of Fk).
In any case: the interface is a thin wrapper, so an extensive test suite is not that required, and we can even postpone to a more consistent tests rework.

### DIFF
--- a/pineappl_py/pineappl/fk_table.py
+++ b/pineappl_py/pineappl/fk_table.py
@@ -24,7 +24,7 @@ class FkTable(PyWrapper):
 
         Parameters
         ----------
-            path : str
+            path : pathlike
                 file path
 
         Returns
--- a/pineappl_py/pineappl/grid.py
+++ b/pineappl_py/pineappl/grid.py
@@ -291,7 +291,7 @@ class Grid(PyWrapper):
 
         Parameters
         ----------
-            path : str
+            path : pathlike
                 file path
 
         Returns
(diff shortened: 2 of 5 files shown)


cschwan:
Looks good to me!

cschwan (approved):

cschwan:
> Speaking of tests: maybe it might be a good chance to extend the IO related ones a bit (`write_lz4` is not tested, nor `read`, `write` and `write_lz4` for FkTables... actually almost nothing is tested of Fk).
> In any case: the interface is a thin wrapper, so an extensive test suite is not that required, and we can even postpone to a more consistent tests rework.

Agreed, I've opened Issue https://github.com/N3PDF/pineappl/issues/108 for that. Thanks for your work on this Issue @AleCandido!

PR #106, opened 2022-01-29
Lists to numpy arrays in python interface

### LINKED ISSUE #82: Return numpy arrays instead of lists in Python interface
At the moment, we're not consistently dealing with arrays/lists in python, since we only tried late the `rust_numpy` crate.

We should then move the formerly implemented methods that returns `Vec<T>` (handled in python as a list) to return whatever `PyArray` is suitable from the mentioned crate.

New with `rust_numpy`:

https://github.com/N3PDF/pineappl/blob/a1fe0350b00a1129e35d680c5dfa4448a59f8fa6/pineappl_py/src/grid.rs#L410

old with `Vec<T>`:

https://github.com/N3PDF/pineappl/blob/a1fe0350b00a1129e35d680c5dfa4448a59f8fa6/pineappl_py/src/grid.rs#L427

https://github.com/N3PDF/pineappl/blob/a1fe0350b00a1129e35d680c5dfa4448a59f8fa6/pineappl_py/src/grid.rs#L444

AleCandido:
This should close #82 

- [x] `import_only_subgrid`
- [x] `fk_table`
- [x] `bin`
- [x] update pure python package for breaking changes ([`convolute_eko`](https://github.com/N3PDF/pineappl/blob/3edf25ee8540a0141bb43425d47a24939fa21634/pineappl_py/src/grid.rs#L332))
- [x] update docs (lists are specified, should be arrays)
- [x] fix tests

I don't believe there is an automatic list to array conversion, so we might need to update also some other code after this (hopefully most is done in the pure python layer).

For this reason, I'm not using arrays where lists are expected. E.g. list of tuples won't be generate as 2 dimensional arrays in most cases:
https://github.com/N3PDF/pineappl/blob/3edf25ee8540a0141bb43425d47a24939fa21634/pineappl_py/src/grid.rs#L275
or lists of objects:
https://github.com/N3PDF/pineappl/blob/3edf25ee8540a0141bb43425d47a24939fa21634/pineappl_py/src/grid.rs#L69

I broke backward compatibility to pass an actual 5 dimensional array as operator, instead of splitting it in flattened + shape.

Most likely we want other updates before merging:
- `Vec`s were used not only in the interface, but even internally, like
  https://github.com/N3PDF/pineappl/blob/c0144e4505fda1f4a835d3a0428b4acefe371197/pineappl/src/grid.rs#L355-L362
  but since we're getting them from python as arrays, and in general arrays are more powerful and suitable for algebra, we might want to keep them as arrays, instead of the ugly conversion that I put at the moment
  https://github.com/N3PDF/pineappl/blob/3edf25ee8540a0141bb43425d47a24939fa21634/pineappl_py/src/grid.rs#L336-L349

### DIFF
--- a/pineappl_py/pineappl/grid.py
+++ b/pineappl_py/pineappl/grid.py
@@ -75,14 +75,14 @@ class Grid(PyWrapper):
                 List of active luminosities
             orders: list(Order)
                 List of available orders
-            bin_limits: BinRemapper
-                bins
+            bin_limits: sequence(float)
+                Bin limits
             subgrid_params : SubgridParams
                 subgrid parameters
         """
-        lumi = [l.raw for l in lumi]
+        lumi = [lentry.raw for lentry in lumi]
         orders = [o.raw for o in orders]
-        return cls(PyGrid(lumi, orders, bin_limits, subgrid_params.raw))
+        return cls(PyGrid(lumi, orders, np.array(bin_limits), subgrid_params.raw))
 
     def subgrid(self, order, bin_, lumi):
         """
@@ -211,13 +211,13 @@ class Grid(PyWrapper):
                 lhapdf like callable with arguments `pid, x, Q2` returning x*pdf for :math:`x`-grid
             alphas : callable
                 lhapdf like callable with arguments `Q2` returning :math:`\alpha_s`
-            order_mask : list(bool)
+            order_mask : sequence(bool)
                 Mask for selecting specific orders. The value `True` means the corresponding order
                 is included. An empty list corresponds to all orders being enabled.
-            bin_indices : list(int)
+            bin_indices : sequence(int)
                 A list with the indices of the corresponding bins that should be calculated. An
                 empty list means that all orders should be calculated.
-            lumi_mask : list(bool)
+            lumi_mask : sequence(bool)
                 Mask for selecting specific luminosity channels. The value `True` means the
                 corresponding channel is included. An empty list corresponds to all channels being
                 enabled.
@@ -236,7 +236,13 @@ class Grid(PyWrapper):
 
         """
         return self.raw.convolute_with_one(
-            pdg_id, xfx, alphas, order_mask, bin_indices, lumi_mask, xi
+            pdg_id,
+            xfx,
+            alphas,
+            np.array(order_mask),
+            np.array(bin_indices),
+            np.array(lumi_mask),
+            xi,
         )
 
     def convolute_eko(self, operators, lumi_id_types="pdg_mc_ids", order_mask=()):
@@ -268,7 +274,7 @@ class Grid(PyWrapper):
         alphas_values = [op["alphas"] for op in operators["Q2grid"].values()]
         return FkTable(
             self.raw.convolute_eko(
-                np.array(operators["q2_ref"]),
+                operators["q2_ref"],
                 np.array(alphas_values),
                 np.array(operators["targetpids"], dtype=np.int32),
                 np.array(operators["targetgrid"]),
@@ -299,3 +305,16 @@ class Grid(PyWrapper):
                 grid object
         """
         return cls(PyGrid.read(path))
+
+    def delete_bins(self, bin_indices):
+        """Delete bins.
+
+        Repeated bins and those exceeding length are ignored.
+
+        Parameters
+        ----------
+        bin_indices : sequence(int)
+            list of indices of bins to removed
+
+        """
+        self.raw.delete_bins(np.array(bin_indices))
--- a/pineappl_py/pineappl/grid.py
+++ b/pineappl_py/pineappl/grid.py
@@ -195,9 +195,9 @@ class Grid(PyWrapper):
         pdg_id,
         xfx,
         alphas,
-        order_mask=(),
-        bin_indices=(),
-        lumi_mask=(),
+        order_mask=np.array([], dtype=bool),
+        bin_indices=np.array([], dtype=np.uint64),
+        lumi_mask=np.array([], dtype=bool),
         xi=((1.0, 1.0),),
     ):
         """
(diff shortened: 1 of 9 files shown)


AleCandido:
Current version should be breaking even tests (even though we have a limited amount of them), and for sure it will break some usage (since we now require numpy arrays everywhere in the input).

We can decide to wrap all functions in the pure python layer, in order to accept both lists and numpy arrays (`np.array` is idempotent), but I don't know if it's worth. What do you think about this @cschwan?

In any case, `pineappl` (and especially interface) is known to be unstable, so it should be fine anyhow.

AleCandido:
P.S.: if anyone (i.e. @cschwan or @felixhekhorn) wants to start a review, start grepping for `Vec` in `pineappl_py/src`, and consider if what is still left behind is find to keep as a `Vec` (there are some cases in which numpy array conversion is possible, like lumi functions, but not being a true linear algebra object I preferred to keep as a `Vec`)

cschwan:
> We can decide to wrap all functions in the pure python layer, in order to accept both lists and numpy arrays (`np.array` is idempotent), but I don't know if it's worth. What do you think about this @cschwan?

Since you're the sole user of the Python interface for the time being, you should decide this. What's more pythonic?

cschwan:
> Current version should be breaking even tests (even though we have a limited amount of them), and for sure it will break some usage (since we now require numpy arrays everywhere in the input).

Are the Python tests being run in the CI?

AleCandido:
> Are the Python tests being run in the CI?

No (even though it's installing python for an unknown reason).

1. should I add to the same workflow?
2. I'd propose to add a workflow running the examples, to keep checking them (at least while releasing); citing rust book 

> Nothing is better than documentation with examples. But nothing is worse than examples that don’t work because the code has changed since the documentation was written.

(section 14.2)

P.S.: do we want to do 1. in this PR or in another one?

AleCandido:
> Since you're the sole user of the Python interface for the time being, you should decide this. What's more pythonic?

Python has no strict type system (only type hints for static analyzers), so the pythonic way is duck-typing: if you need a feature on your input (comparison, arithmetic, ...) check for the feature, not for the type. So everything that can work should do it (but there is the maintainability price, since every time you write a function you'll do it twice, even though it's a thin layer).

cschwan:
> P.S.: do we want to do 1. in this PR or in another one?

I'll let you decide this, I don't have strong opinion about either one.

AleCandido:
Done in 9d2a603. Ready to merge.

cschwan:
Thanks working on this Issue!

pineappl_py/pineappl/grid.py
```diff
@@ -268,7 +274,7 @@ def convolute_eko(self, operators, lumi_id_types="pdg_mc_ids", order_mask=()):
         alphas_values = [op["alphas"] for op in operators["Q2grid"].values()]
         return FkTable(
             self.raw.convolute_eko(
-                np.array(operators["q2_ref"]),
+                operators["q2_ref"],
                 np.array(alphas_values),
```
felixhekhorn, line 278: I wonder where to cast:
- for `targetpids` you're doing it here below, but
- for `alphas_values` you're not doing it
- however it will eventually happen as Rust is requiring it here https://github.com/N3PDF/pineappl/blob/6806ad0f74ef31ee8cfdc27fbc96dea5f6dded66/pineappl_py/src/grid.rs#L330 (of course due to https://github.com/N3PDF/pineappl/blob/6806ad0f74ef31ee8cfdc27fbc96dea5f6dded66/pineappl/src/grid.rs#L371)

this is only one example out of many

felixhekhorn:
With the numpy-fication happening in the interface this is still backward compatible, correct?

AleCandido:
> With the numpy-fication happening in the interface this is still backward compatible, correct?

Yes, that was half of the purpose. The other half being to always support structures that contains the same information.

pineappl_py/pineappl/grid.py
felixhekhorn: > You're also mixing two different functions, `convolute_with_one` and `convolute_eko`, so I'm pretty confused...

Mmm, I think, I pointed to  `convolute_eko` always - in case I didn't sorry :upside_down_face: - as said the thing is general

> I didn't understand what do you mean by _where to cast_.

I mean the difference between this line:
https://github.com/N3PDF/pineappl/blob/6806ad0f74ef31ee8cfdc27fbc96dea5f6dded66/pineappl_py/pineappl/grid.py#L279
- with an explicit cast to int
- so numpy will cast here
- and e.g. `[1.,2.]` will result into `np.array([1,2])`

and this line:
https://github.com/N3PDF/pineappl/blob/6806ad0f74ef31ee8cfdc27fbc96dea5f6dded66/pineappl_py/pineappl/grid.py#L278
- with *no* explict cast here
- so Rust will cast upon consumption (I guess - since I don't know how PyO3 is working)
- as it is requested here https://github.com/N3PDF/pineappl/blob/6806ad0f74ef31ee8cfdc27fbc96dea5f6dded66/pineappl_py/src/grid.rs#L330

The concern is not about passing lists or array, but about the list content type - in Rust language: `PyReadonlyArray1<f64>`

Explicit suggestion: we may change the line to `np.array(alphas_values, dtype=np.float)`

AleCandido: I only have to cast to integers, because they are float by default

felixhekhorn: I don't believe so: https://numpy.org/doc/stable/reference/generated/numpy.array.html?highlight=array#numpy.array
"The desired data-type for the array. If not given, then the type will be determined as the minimum type required to hold the objects in the sequence."

AleCandido: That is: you need only one float to make it float

- alphas is a positive number smaller than one
- same goes for x

AleCandido: P.S.: @felixhekhorn the moment you have a reasonable example breaking, I'll be happy to cast immediately for that one :)

felixhekhorn: I don't - I'd just thought about following the general rule: "I'd rather cast myself then at a random place" ... fine, I accept

felixhekhorn:
Re casting: https://github.com/N3PDF/pineappl/commit/9b4d75bcc900cd6e735bf451a4d21f81f18b2481

AleCandido:
Still in the same direction: float - > int - > uint 🙃

felixhekhorn:
Well, as said you need the explicit cast
```
In [2]: np.array([1,2,3]).dtype
Out[2]: dtype('int64')

In [3]: np.array([1,2,3],dtype=np.uint).dtype
Out[3]: dtype('uint64')
```
otherwise rust is complaining (and not doing anything by the way)

felixhekhorn:
I believe another side-effect (slightly breaking backward compatibility) is https://github.com/NNPDF/fktables/commit/7ab2fe663090a0435734e0f0d8ffd0415854f8e8

Unfortunately I don't have the error message still at hand, but it was saying smth like "output array is read-only" ...

PR #167, opened 2022-08-18
`convolute_eko`: `x2_grid` was determined though not needed

felixhekhorn:
During the debug of https://github.com/NNPDF/runcards/pull/150 we realized there is a bug in PineAPPL `convolute_eko`:
`x2_grid` was determined although it was never needed (because x2 is trivial)

This fix adds a protection to the generation same as there is already for reading: https://github.com/N3PDF/pineappl/blob/04650d99dd52f7985ba4fb2a976980b2b91ce6e7/pineappl/src/grid.rs#L1632

the quickfix (applied in https://github.com/NNPDF/runcards/pull/150/commits/3a5b7c9dc053707bb74d181c69b5add50c5a385f ) is to point the trivial point to an existing point in the x1 grid - this is done also for DIS (and everything derived from that), specificly x2=1, which, by chance, is in the x1 grid; that's also the reason why it was not discovered before

there might be an other bug related: @scarlehoff @andreab1997  I couldn't trigger again the core dump - can you remind me, what do I need to do?

### DIFF
--- a/pineappl/src/grid.rs
+++ b/pineappl/src/grid.rs
@@ -1535,30 +1535,38 @@ impl Grid {
             .find(|subgrid| !subgrid.is_empty())
             .unwrap_or_else(|| unreachable!());
         // source x1/x2 grid might differ and be differently sorted than the operator
-        let x1_grid: Vec<_> = first_non_zero_grid
-            .x1_grid()
-            .iter()
-            .map(|x| {
-                eko_info
-                    .grid_axes
-                    .x_grid
-                    .iter()
-                    .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
-                    .unwrap_or_else(|| unreachable!())
-            })
-            .collect();
-        let x2_grid: Vec<_> = first_non_zero_grid
-            .x2_grid()
-            .iter()
-            .map(|x| {
-                eko_info
-                    .grid_axes
-                    .x_grid
-                    .iter()
-                    .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
-                    .unwrap_or_else(|| unreachable!())
-            })
-            .collect();
+        let x1_grid: Vec<_> = if has_pdf1 {
+            first_non_zero_grid
+                .x1_grid()
+                .iter()
+                .map(|x| {
+                    eko_info
+                        .grid_axes
+                        .x_grid
+                        .iter()
+                        .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
+                        .unwrap_or_else(|| unreachable!())
+                })
+                .collect()
+        } else {
+            Vec::<usize>::new()
+        };
+        let x2_grid: Vec<_> = if has_pdf2 {
+            first_non_zero_grid
+                .x2_grid()
+                .iter()
+                .map(|x| {
+                    eko_info
+                        .grid_axes
+                        .x_grid
+                        .iter()
+                        .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
+                        .unwrap_or_else(|| unreachable!())
+                })
+                .collect()
+        } else {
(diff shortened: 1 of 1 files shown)


andreab1997:
> During the debug of [NNPDF/runcards#150](https://github.com/NNPDF/runcards/pull/150) we realized there is a bug in PineAPPL `convolute_eko`: `x2_grid` was determined although it was never needed (because x2 is trivial)
> 
> This fix adds a protection to the generation same as there is already for reading:
> 
> https://github.com/N3PDF/pineappl/blob/04650d99dd52f7985ba4fb2a976980b2b91ce6e7/pineappl/src/grid.rs#L1632
> 
> the quickfix (applied in [NNPDF/runcards@3a5b7c9](https://github.com/NNPDF/runcards/commit/3a5b7c9dc053707bb74d181c69b5add50c5a385f) ) is to point the trivial point to an existing point in the x1 grid - this is done also for DIS (and everything derived from that), specificly x2=1, which, by chance, is in the x1 grid; that's also the reason why it was not discovered before
> 
> there might be an other bug related: @scarlehoff @andreab1997 I couldn't trigger again the core dump - can you remind me, what do I need to do?

You just need to compute the fks for the integrability grid. It does not crash during the eko calculation.

felixhekhorn:
> You just need to compute the fks for the integrability grid. It does not crash during the eko calculation.

this is not what I meant - the error (concerning PineAPPL and not [pineko](https://github.com/N3PDF/pineko/pull/41) ) in the fk generation was a "reached 'unreachable' code" (or similar) which is circumvented with this PR (it was the `unreachable!()` right there);

instead we had a specific `... core dump ...` for a specific `pineappl` CLI call, which I can't remember which it was ... (I think it was `convolute`, but I can't reproduce the failing PineAPPL file )

scarlehoff:
> instead we had a specific `... core dump ...` for a specific `pineappl` CLI call, which I can't remember which it was ... (I think it was `convolute`, but I can't reproduce the failing PineAPPL file )

It was `convolute` and it was a grid generated with this first commit https://github.com/NNPDF/runcards/pull/150/commits/de4605a73b54c6b6bf266a5fa7ff82c9132120ec however if it is not reproduced it is a good thing.

(I was changing the runcard trying to make it work so that grid in particular might have been fundamentally broken, it might be one of the first one I sent you but boh)

felixhekhorn:
Actually - also all the grids in https://github.com/NNPDF/runcards/pull/132 suffer from this bug because [the xgrid](https://github.com/NNPDF/runcards/blob/54d4a55d171fe068b01eff715e0249f81100a469/pinecards/NNPDF_POS_UP_40/positivity.yaml#L5-L25) does not contain [the trivial x2 point at x2=1](https://github.com/NNPDF/runcards/blob/54d4a55d171fe068b01eff715e0249f81100a469/pinefarm/external/positivity.py#L68)

giacomomagni:
Hi, is it possible to merge this PR? I need this fix to recompute some positivity fk tables at N3LO.

cschwan:
I'm happy to merge it. @felixhekhorn is there something else needed here?

felixhekhorn:
we can merge, I think (if ever I find the other bug, I'll open a new PR)

cschwan (approved):

andreab1997:
It seems that this bug is also affecting the NLO positivities. I will wait this to be merged to compute them :)

felixhekhorn:
To make life for @giacomomagni and @andreab1997 easier: can we also make a release?

giacomomagni:
That will be really appreciated :). And thanks for solving the issue quickly

cschwan:
Here you go: <https://github.com/N3PDF/pineappl/releases/tag/v0.5.5>. All the Python and Rust packages are also released, the conda package releases itself automatically as well, but it usually takes a bit longer.

PR #176, opened 2022-09-16
Distribute CLI as a Python package

AleCandido:
The goal of this PR is just to add metadata for Python packaging.
No new crate/folder is going to be created, because no further code should be added.

To do:
- [x] add required metadata
- [ ] prepare CI for release
- [ ] add [dependency](https://github.com/N3PDF/pineappl/pull/176#issuecomment-1249179607) on this new package in the `pineappl` Python package

In order to test locally, you can use the usual:
- `maturin develop` to build and install in current environment
- `maturin build` to make the actual Python package, without installation

Eventually, we will use `maturin publish` (possibly in the same workflow distributing `pineappl_py`), and just upload the further package `pineappl-cli` to PyPI, just wrapping the Rust binary.
https://github.com/N3PDF/pineappl/blob/81dd719f62103895be245251af38582d05dca80b/.github/workflows/wheels.yml#L20
https://github.com/N3PDF/pineappl/blob/81dd719f62103895be245251af38582d05dca80b/.github/workflows/wheels.yml#L47

### DIFF
--- a/pineappl_cli/pyproject.toml
+++ b/pineappl_cli/pyproject.toml
@@ -1,0 +1,29 @@
+[build-system]
+requires = ["maturin>=0.12,<0.13"]
+build-backend = "maturin"
+
+[project]
+name = "pineappl-cli"
+requires-python = ">=3.6"
+classifiers = [
+  "Development Status :: 4 - Beta",
+  "Intended Audience :: Science/Research",
+  "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
+  "Natural Language :: English",
+  "Operating System :: MacOS",
+  "Operating System :: Microsoft :: Windows",
+  "Operating System :: POSIX",
+  "Programming Language :: Rust",
+  "Topic :: Scientific/Engineering :: Physics",
+]
+
+dependencies = []
+
+[project.urls]
+homepage = "https://n3pdf.github.io/pineappl/"
+documentation = "https://pineappl.readthedocs.io/"
+changelog = "https://github.com/N3PDF/pineappl/blob/master/CHANGELOG.md"
+
+[tool.maturin]
+bindings = "bin"
+strip = true


cschwan:
That's ingenious :+1:! We have to decide which features to activate; right now I suppose it uses the default features, which do not include the importers. However, supporting this might be difficult, unless we link everything together statically (which might be a good idea).

cschwan:
I just compiled it and it seems it ships LHAPDF!

cschwan:
@AleCandido how do I install the binary after having it built with `maturin build`?

AleCandido:
> @AleCandido how do I install the binary after having it built with `maturin build`?

I believe you just want something like [this](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-from-local-archives):
```sh
pip install ./downloads/SomeProject-1.0.4.tar.gz
```

AleCandido:
> I just compiled it and it seems it ships LHAPDF!

Yes, but you know [where it is coming from](https://github.com/cschwan/lhapdf). It is simply a build time dependency, and then Rust compiles everything statically by default.

AleCandido:
Eventually, I would make this a dependency of the `pineappl` Python package.

Logically, it would be the other way round, but I believe to be easier to say: "Just do:
```sh
pip install pineappl
```
and you will get Python API + CLI out of the box".

I would not try instead to make a single bundle. I'm very happy about what `maturin` is providing us right now, so I prefer not to be too demanding.

cschwan:
> Yes, but you know [where it is coming from](https://github.com/cschwan/lhapdf). It is simply a build time dependency, and then Rust compiles everything statically by default.

If you open the wheel by unzipping it you can see that it ships `libLHAPDF.so`.

AleCandido:
> If you open the wheel by unzipping it you can see that it ships `libLHAPDF.so`.

Ah, this is completely unexpected to me.
Most likely depends on the way you packaged it, but it still unexpected.

cschwan:
> Eventually, I would make this a dependency of the `pineappl` Python package.
> 
> Logically, it would be the other way round, but I believe to be easier to say: "Just do:
> 
> ```shell
> pip install pineappl
> ```
> 
> and you will get Python API + CLI out of the box".
> 
> I would not try instead to make a single bundle. I'm very happy about what `maturin` is providing us right now, so I prefer not to be too demanding.

I agree with everything you said.

cschwan:
This looks good. We should build the CLI with all features and release mode,

```
maturin build --release --all-features
```

and ideally we should also install ROOT to support the conversion of `.root` APPLgrids. However, without a custom container that might be impossible.

AleCandido:
> and ideally we should also install ROOT to support the conversion of `.root` APPLgrids. However, without a custom container that might be impossible.

What about a [pre-compiled distribution](https://root.cern/install/#download-a-pre-compiled-binary-distribution)?

In one case I actually have control over the container (and being Linux should be simple enough to have compatible binaries).
In the other two cases we can do it as part of the job, so if binaries are available should be simple enough.

cschwan (approved):

AleCandido:
I merge this, since it's already working, and I'll manually release the first package.
I will introduce the CI support and the `pineappl` package dependency in a later PR.

PR #178, opened 2022-09-16
OO C++ interface

felixhekhorn:
With my collaborators we're trying to [redo a previous analysis](https://inspirehep.net/literature/1940909) and so I was thinking about including a PineAPPL interface into my MC event generator (https://github.com/felixhekhorn/LeProHQ ) and I started to write a proper object oriented wrapper (since I like C++ much better then C ;-) ). (Actually, it will be much faster to just run the code for 1000 replicas then writing the wrapper, but since I like the project I'll try to do it nevertheless - and it can then be used in the future)


felixhekhorn:
Actually, @cschwan there is something I don't understand: how can `pineappl_grid_fill` be used to fill multi-dimensional grids? because my generator computes structure functions and so will be by default two dimensional (x, Q2) - and I can make it easily higher dimensional (say + heavy quark rapidity)

AleCandido:
> + heavy quark rapidity

Bins (the second dimension of the grid, according to `obl` sorting) are actually a multi-index. So, you can stuff in there how many dimension you wish, for as long as they correspond to external (measured) kinematics.

cschwan:
@felixhekhorn basically what @AleCandido said: for n-dimensional distributions you need to map them to 1-dimenional one using essentially bins, and a posteriori you can

cschwan:
Opening/closing: that was me having butterfingers :smile: .

AleCandido:
> I'm not convinced this is the best approach, but for time being that's what we have.

In a sense is the common one: even Numpy and other array libraries they always store one dimensional data, just because the memory model is linear. Usually, multi-dimensionality is achieved in two ways:
- at *data level* (the representation), by using [**strides**](https://numpy.org/doc/stable/glossary.html?highlight=strides#term-stride)
- at the *abstract level* (the represented object), by means of **coordinates**, i.e. you maintain an auxiliary array, telling you the bounds in the represented space

Maybe the bins machinery can be improved, and maybe we can think about a better way to maintain coordinates, possibly leveraging an external array crate like `ndarray`.
But for the grid itself you wanted to manage on your own, because it is already a complicate data structure. The only other option I see is to populate an `NDArray` of "partial grids" (i.e. order, lumi, and subgrids) one for each bin.

felixhekhorn:
> you need to map them to 1-dimenional one using essentially bins, and after having filled the grid you can remap,

okay, I see - and I understand the technical limitations ... but then I wonder whether I should provide some sugar here on this level ... in any case I will first complete the example, then attempt my implementation and if I feel I need more, do more

cschwan:
> [..] and it can then be used in the future)

In the sense of `pinefarm`?

felixhekhorn:
> In the sense of `pinefarm`?

if there would an interest (EIC, maybe? the HERA stuff is mostly already in as combined), why not ... (and keep in mind for the *fully inclusive* case yadism is already using my code under the hood) otherwise mostly for me (and my collaborators) and my studies ...

felixhekhorn:
The current implementation reproduces one-to-one the C counter part (now moved into `examples/c`).

Why I believe this is a good idea:
- no more `delete` in the example file, but this is done once and forever in the wrapper
- as a nice side effect, we reduce the file length from 171 line to 153 lines
- names are now nicely scoped: `pineappl_grid_set_key_value` -> `PineAPPL::Grid::set_key_value` ("Namespaces are one honking great idea -- let's do more of those!")
- the interface is actually closer to Rust as it uses e.g. `Order` as a struct

Things that could be improved:
- the interface is not complete, but just makes the example back working - if I'm allowed I'd rather do the completion upon request at a later time; as said the second step is to actually use it and doing so, I'll discover new functions that are needed (e.g. `pineappl_grid_set_remapper`)
- I use `std::list` and `std::vector` in parallel - I usually prefer list since this is the "simplest implementation" as it just requires pointers, however sometimes vector is also nice too have, e.g. for bins for which `at` is quite nice to have; should I do something about this?

AleCandido:
> * the interface is not complete, but just makes the example back working - if I'm allowed I'd rather do the completion upon request at a later time; as said the second step is to actually use it and doing so, I'll discover new functions that are needed (e.g. `pineappl_grid_set_remapper`)

This is exactly what we are doing for the Python wrapper. If it's working there, can work even here :)

> * I use `std::list` and `std::vector` in parallel - I usually prefer list since this is the "simplest implementation" as it just requires pointers, however sometimes vector is also nice too have, e.g. for bins for which `at` is quite nice to have; should I do something about this?

This I would not do: according to Stroustroup (I read it in [A tour of C++](https://www.stroustrup.com/tour2.html), that should actual be just a part of [The C++ Programming Language](https://www.stroustrup.com/4th.html)) always use a `std::vector` unless you have a very good reason not to do it.
E.g., if you want to implement a dequeue, it is much better to use a `std::list` (most of the times at least). As well as if you have to insert many elements in the middle, and stuffs like that.
But if you are not 90-100% convinced that your object has to be something else, than use `std::vector`, it is by far the most efficient.

cschwan:
In almost all cases `std::vector<T>` is preferable because it lays out its elements subsequently in memory and so you can easily interface with C by using the `data` member function.

felixhekhorn:
> In almost all cases `std::vector<T>` is preferable because it lays out its elements subsequently in memory and so you can easily interface with C by using the `data` member function.

that's why I used `vector` some times :innocent: - however, everything moved to `vector` in https://github.com/N3PDF/pineappl/pull/178/commits/16c5b7f193df4dd46cd49e473a541f4850073cb3

AleCandido:
> In almost all cases `std::vector<T>` is preferable because it lays out its elements subsequently in memory and so you can easily interface with C by using the `data` member function.

That might be one reason, but the most quoted one is because of memory management and operations: in practice `std::vector` has the best heuristic to limit the amount of allocations, and it has O(1) access performance.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    size_t count() const { return pineappl_lumi_count(this->raw); }
+
+    /**
+     * @brief add a luminosity function
+     * @param c luminosity function
+     */
+    void add(const std::vector<LumiEntry>& c) const {
+        size_t n = 0;
+        int32_t *pids = new int32_t;
+        double *weights = new double;
```
cschwan, line 80: That's a memory leak. Use `std::vector<double>` instead.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+        for (auto it = c.cbegin(); it != c.cend(); ++it, ++n) {
```
cschwan, line 81: Use `std::transform` from `<algorithm>` with a lambda.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    /** @brief number of bins */
+    size_t n_bins = 0;
+
+    /** @brief number of lumis */
+    size_t n_lumis = 0;
+
+public:
+
+    /** @brief underlying raw object */
+    pineappl_grid *raw;
```
cschwan, line 126: Make it `private`.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+     * @brief constructor
+     * @param lumi luminosity
+     * @param orders orders
+     * @param bins bins
+     * @param key_val additional informations
+     */
+    Grid(const Lumi& lumi, const std::vector<Order>& orders, const std::vector<double>& bins, const KeyVal& key_val) {
+        this->n_lumis = lumi.count();
+        // cast orders
+        uint32_t *raw_orders = new uint32_t[4 * orders.size()];
```
cschwan, line 138: Use a `std::vector<std::unint32_t>` instead of an array.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+        for (auto it = orders.cbegin(); it != orders.cend(); ++it, ++this->n_orders){
+            raw_orders[4*this->n_orders + 0] = it->alphas;
+            raw_orders[4*this->n_orders + 1] = it->alpha;
+            raw_orders[4*this->n_orders + 2] = it->logxir;
+            raw_orders[4*this->n_orders + 3] = it->logxif;
+        }
+        //this->n_bins = bins.size();
+        // cast bins
+        double *raw_bins = new double[bins.size()];
```
cschwan, line 147: Delete this line ...

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+        for (auto it = bins.cbegin(); it != bins.cend(); ++it, ++this->n_bins)
```
cschwan, line 148: and this one ...

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+            raw_bins[this->n_bins] = *it;
```
cschwan, line 149: and this ...

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+        // now, we got all we need
+        this->raw = pineappl_grid_new(lumi.raw, this->n_orders, raw_orders, this->n_bins - 1, raw_bins, key_val.raw);
```
cschwan, line 151: and pass `bins.data()` instead of `raw_bins`.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    pineappl_grid *raw;
+
+    /**
+     * @brief constructor
+     * @param lumi luminosity
+     * @param orders orders
+     * @param bins bins
+     * @param key_val additional informations
+     */
+    Grid(const Lumi& lumi, const std::vector<Order>& orders, const std::vector<double>& bins, const KeyVal& key_val) {
```
cschwan, line 135: Rename `bins` to `bin_limits`.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * @brief perform a convolution of the grid with PDFs
+     * @param pdg_id particle ID
+     * @param pdf PDF
+     * @param xi_ren renormalization scale variation
+     * @param xi_fac factorization scale variation
+     * @return prediction for each bin
+     * @return prediction for each bin
+     */
+    std::vector<double> convolute_with_one(int32_t pdg_id, LHAPDF::PDF *pdf, double xi_ren = 1.0, double xi_fac = 1.0 ) const {
```
cschwan, line 183: Remove this member function ...

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+     * @brief perform a convolution of the grid with PDFs
+     * @param pdg_id particle ID
+     * @param pdf PDF
+     * @param order_mask order mask
+     * @param lumi_mask luminosity mask
+     * @param xi_ren renormalization scale variation
+     * @param xi_fac factorization scale variation
+     * @return prediction for each bin
+     */
+    std::vector<double> convolute_with_one(int32_t pdg_id, LHAPDF::PDF *pdf, const std::vector<bool>& order_mask, const std::vector<bool>& lumi_mask, double xi_ren = 1.0, double xi_fac = 1.0) const {
```
cschwan, line 199: and give `order_mask` and `lumi_mask` default parameters (empty vectors). Check if each of them are empty and if they are, pass `nullptr` instead below to `pineappl_grid_convolute_with_one`, which is understood as 'enable everything'.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+        // prepare LHAPDF stuff
+        auto xfx = [](int32_t id, double x, double q2, void* pdf) {
+            return static_cast <LHAPDF::PDF*> (pdf)->xfxQ2(id, x, q2);
+        };
+        auto alphas = [](double q2, void* pdf) {
+            return static_cast <LHAPDF::PDF*> (pdf)->alphasQ2(q2);
+        };
+        // cast order_mask
+        bool *raw_order_mask = new bool[this->n_orders];
```
cschwan, line 208: Delete these lines and pass `order_mask.data()` and `lumi_mask.data()` instead to the CAPI.

examples/cpp/dyaa.cpp
```diff
@@ -1,11 +1,13 @@
-#include <pineappl_capi.h>
 #include <LHAPDF/LHAPDF.h>
 
 #include <cmath>
 #include <cstddef>
 #include <cstdio>
 #include <random>
 #include <vector>
+#include <list>
```
cschwan, line 8: Delete this header.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
+/**
+ * @file PineAPPL.hpp
+ * @brief object oriented interface to PineAPPL
+ */
+#ifndef PineAPPL_HPP_
+#define PineAPPL_HPP_
+
+#include <string>
+#include <list>
```
cschwan, line 9: Delete this header.

examples/cpp/dyaa.cpp
```diff
@@ -55,7 +57,7 @@ Psp2to2 hadronic_pspgen(std::mt19937& rng, double mmin, double mmax) {
     return { s, t, u, x1, x2, jacobian };
 }
 
-void fill_grid(pineappl_grid* grid, std::size_t calls) {
+void fill_grid(PineAPPL::Grid* grid, std::size_t calls) {
```
cschwan, line 60: Pass a reference instead of a pointer.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+     * @param lumi luminosity
+     * @param orders orders
+     * @param bins bins
+     * @param key_val additional informations
+     */
+    Grid(const Lumi& lumi, const std::vector<Order>& orders, const std::vector<double>& bins, const KeyVal& key_val) {
+        this->n_lumis = lumi.count();
+        // cast orders
+        uint32_t *raw_orders = new uint32_t[4 * orders.size()];
+        for (auto it = orders.cbegin(); it != orders.cend(); ++it, ++this->n_orders){
```
cschwan, line 139: Use a [range-for](https://en.cppreference.com/w/cpp/language/range-for) to iterate over `orders` and assign `this->n_orders` the correct value - this is much more readable than incrementing it up to the correct value.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+#include <list>
+#include <LHAPDF/LHAPDF.h>
+
+#include <pineappl_capi.h>
+
+/** @brief object oriented interface to PineAPPL  */
+namespace PineAPPL {
+
+/** @brief Key-value storage for passing optional information during grid creation */
+struct KeyVal {
```
cschwan, line 18: I suggest to use `struct` for PODs (like `KeyVal` and `Order`) and `class` otherwise.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    void set_key_value(std::string key, std::string value) const {
+        pineappl_grid_set_key_value(this->raw, key.c_str(), value.c_str());
+    }
+
+    /**
+     * @brief Get a metadata entry
+     * @param key key
+     * @return value
+     */
+    std::string get_key_value(std::string key) const {
```
cschwan, line 245: Pass `key` by const reference.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    void write(std::string filename) const {
+        pineappl_grid_write(this->raw, filename.c_str());
+    }
+
+    /**
+     * @brief Set a metadata entry
+     * @param key key
+     * @param value value
+     */
+    void set_key_value(std::string key, std::string value) const {
```
cschwan, line 236: Pass `key` and `value` by const reference.

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+        delete[] raw_order_mask;
+        delete[] raw_lumi_mask;
+        return results;
+    }
+
+    /**
+     * @brief Write grid to file
+     * @param filename file name
+     */
+    void write(std::string filename) const {
```
cschwan, line 227: Pass `filename` by const reference.

cschwan (changes requested):

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    /** @brief Exponent of the logarithm of the scale factor of the renomalization scale. */
+    uint32_t logxir;
+
+    /** @brief Exponent of the logarithm of the scale factor of the factorization scale. */
+    uint32_t logxif;
+};
+
+class Grid {
+    /** @brief number of orders */
+    size_t n_orders = 0;
```
cschwan, line 115: Remove `n_orders`, `n_bins` and `n_lumis` as member variables and instead make them local to `Grid::convolute_with_one`. The problem is that there are operations that could possibly modify them iternally in `raw` and then they are out-of-sync in C++.

cschwan (changes requested):

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+    /** @brief constructor */
+    KeyVal() { this->raw = pineappl_keyval_new(); }
+
+    /** @brief destructor */
+    ~KeyVal() { pineappl_keyval_delete(this->raw); }
+
+    
+    /** @name setter */
+    ///@{
+    void set_double(const std::string& key, const double value) const { pineappl_keyval_set_double(this->raw, key.c_str(), value); }
```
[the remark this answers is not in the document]
felixhekhorn: this is a guard to myself, so I can't unintentionally change its value (and it doesn't hurt) - in fact, I added a few more to be even more strict

examples/cpp/PineAPPL.hpp
felixhekhorn: AFAIK, the only difference between `struct` and `class` is the default scope - since here everything is public I'd like to keep `struct` (thus saving one line)

examples/cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+#include <pineappl_capi.h>
+
+/** @brief object oriented interface to PineAPPL  */
+namespace PineAPPL {
+
+/** @brief Key-value storage for passing optional information during grid creation */
+struct KeyVal {
+    /** @brief underlying raw object */ 
+    pineappl_keyval *raw;
```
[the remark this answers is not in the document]
felixhekhorn: I'd like to keep it public, because
- this way one can always fallback to the C interface (e.g. if I'm missing a function)
- and otherwise I have to make all classes friends with each other (which is unnecessarily complicated)
- and I can keep everything public (see https://github.com/N3PDF/pineappl/pull/178/files#r974458714)

examples/cpp/PineAPPL.hpp
felixhekhorn: https://github.com/N3PDF/pineappl/pull/178/files#r974460355

examples/cpp/PineAPPL.hpp
felixhekhorn: fine - then this becomes also a struct: https://github.com/N3PDF/pineappl/pull/178/files#r974460355

examples/cpp/PineAPPL.hpp
felixhekhorn: With a single loop I can fill both arrays

examples/cpp/PineAPPL.hpp
felixhekhorn: `n_orders` is now local and correct initialized - however, for the loop I need the index since `raw_orders` has more elements

examples/cpp/PineAPPL.hpp
felixhekhorn: I can't - because `PineAPPL.hpp:217:41: error: use of deleted function ‘void std::vector<bool, _Alloc>::data() [with _Alloc = std::allocator<bool>]’
  217 |         auto *raw_mask = order_mask.data();` seems like `vector<bool>` is special and even considered evil ...

examples/cpp/PineAPPL.hpp
felixhekhorn: I would prefer to keep this, because I think it matches a common use case: the variation of scales is much more common than shutting orders or lumis down (I would say). If I'm giving the possibility to overload, let's use it! I'd see it as the cpp approach to keyword arguments (in some sense). However, if you insist, I can drop it ...

AleCandido (commented):
I'm not as experienced as @cschwan in C++, the only thing I can reasonably comment about is the following: you're writing the C++ interface wrapping the C one, but at this point is just sugar for the user, with the same (and possibly more) limitations than the C API.

I would consider making it directly on top of the Rust one, so distributing another crate `pineappl_cppapi`. But this only if you can limit the amount of boilerplate, because if the two folders have too much in common it makes no sense (we would be just duplicating).

AleCandido:
Ah sorry, I also compiled and ran, and I can get a grid.
The only weird thing is that I get practically the same output for c and cpp example, but not the same grid: they differ as binaries (a bit unexpected, but mostly uninformative) and I can not compare them with the CLI, because of the following reason:
```
Error: bins limits differ
```

felixhekhorn:
> Ah sorry, I also compiled and ran, and I can get a grid. The only weird thing is that I get practically the same output for c and cpp example, but not the same grid: they differ as binaries (a bit unexpected, but mostly uninformative) and I can not compare them with the CLI, because of the following reason:
> 
> ```
> Error: bins limits differ
> ```

this is due to float arithmetics - I use an iterator to compute the bins, Christopher a fixed list:
```
$ pineappl diff c/DY-LO-AA.pineappl.lz4 cpp/DY-LO-AA.pineappl.lz4 NNPDF31_nlo_as_0118_luxqed --ignore-bin-limits
LHAPDF 6.4.0 loading /home/felix/local/share/LHAPDF/NNPDF31_nlo_as_0118_luxqed/NNPDF31_nlo_as_0118_luxqed_0000.dat
NNPDF31_nlo_as_0118_luxqed PDF set, member #0, version 2; LHAPDF ID = 324900
b    x1                O(as^0 a^2)             
--+---+---+------------+------------+----------
 0   0 0.1 5.2631092e-1 5.2631092e-1    0.000e0
 1 0.1 0.2 5.2549083e-1 5.2549083e-1    0.000e0
 2 0.2 0.3 5.2468241e-1 5.2468241e-1  6.661e-16
 3 0.3 0.4 5.1883402e-1 5.1883402e-1 -6.661e-16
 4 0.4 0.5 5.1754819e-1 5.1754819e-1    0.000e0
 5 0.5 0.6 5.0088406e-1 5.0088406e-1    0.000e0
 6 0.6 0.7 4.9053252e-1 4.9053252e-1    0.000e0
 7 0.7 0.8 4.6757338e-1 4.6757338e-1 -1.110e-15
 8 0.8 0.9 4.3931594e-1 4.3931594e-1    0.000e0
 9 0.9   1 3.9929209e-1 3.9929209e-1    0.000e0
10   1 1.1 3.7068008e-1 3.7068008e-1 -9.992e-16
11 1.1 1.2 3.2647167e-1 3.2647167e-1  2.220e-15
12 1.2 1.3 2.8493445e-1 2.8493445e-1    0.000e0
13 1.3 1.4 2.4867234e-1 2.4867234e-1  2.220e-15
14 1.4 1.5 2.1104193e-1 2.1104193e-1    0.000e0
15 1.5 1.6 1.7974387e-1 1.7974387e-1    0.000e0
16 1.6 1.7 1.4714925e-1 1.4714925e-1  2.220e-15
17 1.7 1.8 1.2055655e-1 1.2055655e-1    0.000e0
18 1.8 1.9 9.4916254e-2 9.4916254e-2  2.220e-15
19 1.9   2 7.2557198e-2 7.2557198e-2 -2.109e-15
20   2 2.1 5.0569670e-2 5.0569670e-2    0.000e0
21 2.1 2.2 3.4917881e-2 3.4917881e-2    0.000e0
22 2.2 2.3 1.9675180e-2 1.9675180e-2  4.441e-15
23 2.3 2.4 5.5653059e-3 5.5653059e-3    0.000e0
Thanks for using LHAPDF 6.4.0. Please make sure to cite the paper:
  Eur.Phys.J. C75 (2015) 3, 132  (http://arxiv.org/abs/1412.7420)
```

cschwan:
> I'm not as experienced as @cschwan in C++, the only thing I can reasonably comment about is the following: you're writing the C++ interface wrapping the C one, but at this point is just sugar for the user, with the same (and possibly more) limitations than the C API.
> 
> I would consider making it directly on top of the Rust one, so distributing another crate `pineappl_cppapi`. But this only if you can limit the amount of boilerplate, because if the two folders have too much in common it makes no sense (we would be just duplicating).

I was thinking about making a similar comment early on, but I think having this code for users more familiar with C++ than C is a good idea, they can simply copy it. It's similar to the Fortran module that we also have in an example directory.

cschwan:
> this is due to float arithmetics - I use an iterator to compute the bins, Christopher a fixed list:

This is because we perform floating point comparisons without any tolerance. If you run `cargo clippy` there are numerous warnings, which I intend to fix at some point. They could become really problematic if you think about `Grid::convolute_eko` and the `x`-grid points.

cschwan (changes requested):

examples/cpp/PineAPPL.hpp
felixhekhorn: I did as you said - but now I dislike that C and C++ do not share the very same structure (which could possibly confusing)

felixhekhorn:
Now, I have
```
$ md5sum DY-LO-AA.pineappl.lz4 
22c711a9b85f75936cd3d21d1cab4d11  DY-LO-AA.pineappl.lz4
$ md5sum ../basic-capi-usage/DY-LO-AA.pineappl.lz4 
e3785e3a8f145c5f9a960e259b5e71a3  ../basic-capi-usage/DY-LO-AA.pineappl.lz4
$ pineappl diff DY-LO-AA.pineappl.lz4 ../basic-capi-usage/DY-LO-AA.pineappl.lz4 NNPDF31_nlo_as_0118_luxqed
LHAPDF 6.4.0 loading /home/felix/local/share/LHAPDF/NNPDF31_nlo_as_0118_luxqed/NNPDF31_nlo_as_0118_luxqed_0000.dat
NNPDF31_nlo_as_0118_luxqed PDF set, member #0, version 2; LHAPDF ID = 324900
b    x1               O(as^0 a^2)           
--+---+---+------------+------------+-------
 0   0 0.1 5.2631092e-1 5.2631092e-1 0.000e0
 1 0.1 0.2 5.2549083e-1 5.2549083e-1 0.000e0
 2 0.2 0.3 5.2468241e-1 5.2468241e-1 0.000e0
 3 0.3 0.4 5.1883402e-1 5.1883402e-1 0.000e0
 4 0.4 0.5 5.1754819e-1 5.1754819e-1 0.000e0
 5 0.5 0.6 5.0088406e-1 5.0088406e-1 0.000e0
 6 0.6 0.7 4.9053252e-1 4.9053252e-1 0.000e0
 7 0.7 0.8 4.6757338e-1 4.6757338e-1 0.000e0
 8 0.8 0.9 4.3931594e-1 4.3931594e-1 0.000e0
 9 0.9   1 3.9929209e-1 3.9929209e-1 0.000e0
10   1 1.1 3.7068008e-1 3.7068008e-1 0.000e0
11 1.1 1.2 3.2647167e-1 3.2647167e-1 0.000e0
12 1.2 1.3 2.8493445e-1 2.8493445e-1 0.000e0
13 1.3 1.4 2.4867234e-1 2.4867234e-1 0.000e0
14 1.4 1.5 2.1104193e-1 2.1104193e-1 0.000e0
15 1.5 1.6 1.7974387e-1 1.7974387e-1 0.000e0
16 1.6 1.7 1.4714925e-1 1.4714925e-1 0.000e0
17 1.7 1.8 1.2055655e-1 1.2055655e-1 0.000e0
18 1.8 1.9 9.4916254e-2 9.4916254e-2 0.000e0
19 1.9   2 7.2557198e-2 7.2557198e-2 0.000e0
20   2 2.1 5.0569670e-2 5.0569670e-2 0.000e0
21 2.1 2.2 3.4917881e-2 3.4917881e-2 0.000e0
22 2.2 2.3 1.9675180e-2 1.9675180e-2 0.000e0
23 2.3 2.4 5.5653059e-3 5.5653059e-3 0.000e0
```
so they are not the exact same binary, but for pineappl they are the same

examples/cpp/PineAPPL.hpp
cschwan: That's true, but at least this is something caught at compile time.

cschwan:
@felixhekhorn if you like you can compare the uncompressed grids, they might be the same. It's likely that lz4 saves a date somewhere.

felixhekhorn:
@cschwan I believe to have addressed everything from the review, so I'd be happy to merge - however, it seems I somehow broke the Fortran example in the workflow (local there is no problem) and I've no idea how, do you?

examples/object-oriented-cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,242 @@
...
+     * @brief Returns the number of combinations of the luminosity function `lumi` for the specified entry.
+     * @param entry position in lumi
+     */
+    std::size_t combinations(const std::size_t entry) const { return pineappl_lumi_combinations(this->raw, entry); }
+};
+
+/** @brief Coupling powers for each grid. */
+struct Order {
+    /** @brief Exponent of the strong coupling. */
+    uint32_t alphas;
```
cschwan, line 99: Here are a couple of `std::` missing.

examples/object-oriented-cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,242 @@
...
+     * @brief perform a convolution of the grid with PDFs
+     * @param pdg_id hadron ID
+     * @param pdf PDF
+     * @param xi_ren renormalization scale variation
+     * @param xi_fac factorization scale variation
+     * @param order_mask order mask
+     * @param lumi_mask luminosity mask
+     * @return prediction for each bin
+     */
+    std::vector<double> convolute_with_one(const std::int32_t pdg_id, LHAPDF::PDF* pdf, const double xi_ren = 1.0, const double xi_fac = 1.0, const std::vector<bool>& order_mask = {}, const std::vector<bool>& lumi_mask = {}) const {
```
cschwan, line 179: Pass `pdf` as a reference.

examples/object-oriented-cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,242 @@
...
+        auto xfx = [](std::int32_t id, double x, double q2, void* pdf) {
+            return static_cast <LHAPDF::PDF*> (pdf)->xfxQ2(id, x, q2);
+        };
+        auto alphas = [](double q2, void* pdf) {
+            return static_cast <LHAPDF::PDF*> (pdf)->alphasQ2(q2);
+        };
+        // cast order_mask
+        std::unique_ptr<bool[]> raw_order_mask;
+        if (order_mask.size() > 0) {
+            raw_order_mask = std::unique_ptr<bool[]> (new bool[this->order_count()]);
```
cschwan, line 190: Here you can use `order_mask.size()` instead of `this->order_count()`. If they're not the same you will have a problem anyway.

examples/object-oriented-cpp/PineAPPL.hpp
```diff
@@ -0,0 +1,242 @@
...
+            std::copy(order_mask.begin(), order_mask.end(), &raw_order_mask[0]);
+        }
+        // cast lumi mask
+        pineappl_lumi* l = pineappl_grid_lumi(this->raw);
+        const std::size_t n_lumis = pineappl_lumi_count(l);
+        pineappl_lumi_delete(l);
+        std::unique_ptr<bool[]> raw_lumi_mask;
+        if (lumi_mask.size() > 0) {
+            raw_lumi_mask = std::unique_ptr<bool[]>(new bool[n_lumis]);
```
cschwan, line 199: Same here, use `lumi_mask.size()`.

examples/object-oriented-cpp/dyaa.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    // create the PineAPPL grid with default interpolation and binning parameters
+    PineAPPL::KeyVal kv;
+    PineAPPL::Grid grid(lumi, orders, bins, kv);
+
+    // fill the grid with phase-space points
+    fill_grid(grid, 10000000);
+
+    // perform a convolution of the grid with PDFs
+    auto* pdf = LHAPDF::mkPDF("NNPDF31_nlo_as_0118_luxqed", 0);
```
cschwan, line 126: Capture `pdf` in a `std::unique_ptr`!

cschwan (changes requested):

examples/object-oriented-cpp/PineAPPL.hpp
```diff
@@ -172,27 +172,11 @@ struct Grid {
...
-     * @param pdg_id hadron ID
-     * @param pdf PDF
      * @param order_mask order mask
      * @param lumi_mask luminosity mask
-     * @param xi_ren renormalization scale variation
-     * @param xi_fac factorization scale variation
      * @return prediction for each bin
      */
-    std::vector<double> convolute_with_one(const std::int32_t pdg_id, LHAPDF::PDF* pdf, const std::vector<bool>& order_mask, const std::vector<bool>& lumi_mask, const double xi_ren = 1.0, const double xi_fac = 1.0) const {
+    std::vector<double> convolute_with_one(const std::int32_t pdg_id, LHAPDF::PDF* pdf, const double xi_ren = 1.0, const double xi_fac = 1.0, const std::vector<bool>& order_mask = {}, const std::vector<bool>& lumi_mask = {}) const {
```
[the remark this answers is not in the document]
felixhekhorn: I guess this is double with the one below ...

examples/object-oriented-cpp/PineAPPL.hpp
felixhekhorn: I did as requested - but I don't like it much: `pdf` is a pointer in `dyaa.cpp` since lhapdf natively delivers it like this, then I cast it to reference inside `convolute_with_one` just to pass it in the end as pointer to the CAPI; why can't it be pointer all the way (as the beginning and the end request)? Agreed, pointers are potentially dangerous but they are also handy sometimes ... and there is no complicated structure here (it is owned by `dyaa.cpp` from the beginning to the end)

cschwan: From the API point-of-view a reference more clearly expresses that a reference to a valid `PDF` object is required, whereas pointers can be `nullptr`, which could be allowed. This isn't the case here, so a reference should be accepted and everything follows from there.

examples/object-oriented-cpp/PineAPPL.hpp
```diff
@@ -226,7 +251,7 @@ struct Grid {
      */
     std::string get_key_value(const std::string &key) const {
         auto* value = pineappl_grid_key_value(this->raw, key.c_str());
-        const std::string res(value);
+        std::string res(value);
```
cschwan, line 254: According to `clang-tidy` the modifier `const` prevents a returned value optimization.

examples/object-oriented-cpp/PineAPPL.hpp
```diff
@@ -22,6 +21,14 @@ struct KeyVal {
...
     KeyVal() : raw(pineappl_keyval_new()) {}
 
+    KeyVal(const KeyVal&) = delete;
+
+    KeyVal(KeyVal&&) = delete;
+
+    KeyVal& operator=(const KeyVal&) = delete;
+
+    KeyVal& operator=(KeyVal&&) = delete;
+
```
cschwan, line 31: Since we have a pointer as a member all other constructors and assignment operators should be deleted; otherwise you can copy the C++ object that doesn't really copy `raw`.

cschwan:
@felixhekhorn have a look at it again. If you have `clang-format` could you please try to run over it? Otherwise it can be merged now!

examples/object-oriented-cpp/Makefile
```diff
@@ -0,0 +1,12 @@
+CXX = c++ -std=c++11
```
AleCandido, line 1: Usually this is just the binary, flags you put in `CXXFLAGS`

cschwan: That depends. The flag effectively changes the compiler, so it should go there. See also the [autoconf macro](https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html#ax_cxx_compile_stdcxx_11).

felixhekhorn: So should I revert?

cschwan: It probably doesn't matter at this point.

felixhekhorn:
Apart from stupid stuff, [cpplint](https://github.com/cpplint/cpplint) complains about:
```
PineAPPL.hpp:10:  Found C system header after C++ system header. Should be: PineAPPL.h, c system, c++ system, other.  [build/include_order] [4]
PineAPPL.hpp:231:  Is this a non-const reference? If so, make const or use a pointer: LHAPDF::PDF &pdf  [runtime/references] [2]
```
should I do something?

pineappl_capi/src/lib.rs
```diff
@@ -16,7 +16,7 @@
 //! on the github page.
 //!
 //! [`PineAPPL`]: https://docs.rs/pineappl
-//! [C++ example]: https://github.com/N3PDF/pineappl/tree/master/examples/cpp
+//! [C example]: https://github.com/N3PDF/pineappl/tree/master/examples/basic-capi-usage
```
cschwan, line 19: Please revert this, because the example is written in C++.

pineappl_capi/src/lib.rs
```diff
@@ -7,7 +7,7 @@
 //! functions that allow one to use [`PineAPPL`] without having to write Rust code, and instead
 //! using
 //!
-//! * C or C++ (see the [C++ example] in the repository).
+//! * C or C++ (see the [C example] in the repository).
```
cschwan, line 10: See below.

examples/object-oriented-cpp/Makefile
felixhekhorn: Fine

cschwan (changes requested):

cschwan:
> Apart from stupid stuff, [cpplint](https://github.com/cpplint/cpplint) complains about:
> 
> ```
> PineAPPL.hpp:10:  Found C system header after C++ system header. Should be: PineAPPL.h, c system, c++ system, other.  [build/include_order] [4]
> PineAPPL.hpp:231:  Is this a non-const reference? If so, make const or use a pointer: LHAPDF::PDF &pdf  [runtime/references] [2]
> ```
> 
> should I do something?

The first one you can follow, what's the reasoning for the second one?

cschwan:
I'll leave you the choice; I consider it good enough as is.

PR #179, opened 2022-09-20
Distribute Python Packages

### LINKED ISSUE #177: Python Distribution
Follow up of #176 

- [ ] prepare CI for release
- [ ] add [dependency](https://github.com/N3PDF/pineappl/pull/176#issuecomment-1249179607) on this new package in the `pineappl` Python package

The overall idea is to build everything in the CI, and supplement the current default PineAPPL package with CLI installation.

This partly address #129, since a quick binary installation will be available through `pip` (but it does not solve the problem of the C API).

alecandido:
Closes #177 

I already updated the container (notice that I remove the default entrypoint, now it runs interactively with bash), and the wheels releasing workflow.

Still, we have to:
- [ ] fix the missing symbol issue
- [ ] check the CI is actually working (non-trivial, we will just try at some point)
- [ ] update dependencies
     - i.e. make `pineappl` depedn on `pineappl_cli`, see #177

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -12,12 +12,28 @@ jobs:
     container:
       image: docker://ghcr.io/n3pdf/maturin:0.2.1
     steps:
+      - name: Set Path
+        run: |
+          echo "PATH=${HOME}/local/bin:${HOME}/.cargo/bin:${PATH}" >> $GITHUB_ENV
+      - name: Clone repository
+        run: |
+          cd ${HOME}
+          # clone pineappl code
+          git clone https://github.com/N3PDF/pineappl.git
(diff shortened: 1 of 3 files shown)


cschwan:
In case this is a problem with pip @AleCandido opened an Issue here: <https://github.com/pypa/auditwheel/issues/396>.

AleCandido:
> In case this is a problem with pip @AleCandido opened an Issue here: [pypa/auditwheel#396](https://github.com/pypa/auditwheel/issues/396).

Marked as duplicated of https://github.com/pypa/auditwheel/issues/340

Since that issue suggests a workaround, I opened https://github.com/PyO3/maturin/issues/1182 to ask if it's possible to implement it in `maturin`.

PR #181, opened 2022-09-29
Remove leftover ImportOnlySubgridV1

felixhekhorn:
this actually broke `convolute_eko` - not sure this is the only thing to do, but at least it is running again ...


cschwan:
Yes, I ran into the same bug.

cschwan (approved):
Merge when the CI gives the OK.

Issue #151, opened 2022-07-02
Remove empty unused `x1` and `x2` grid values in subgrids


cschwan:
We already remove unused grid points in the `muf2`/`mur2` dimension, and we should also be able to do that with `x1` and `x2`. This might require some modifications here and there, but should improve the size of each subgrid as they usually have a small range in x in which the subgrid is non-zero.

This will require changing `Grid::optimize` to also modify `ImportOnlySubgridV1` subgrids.

cschwan:
@AleCandido @felixhekhorn I am working on this change; it will not decrease the file sizes, but I hope it will improve the performance of the evolution a bit. The price that we have to pay for this is that instead of having always the same x-grid everywhere - the usual 50 points - the grids return a *subset* of these 50 points. So while the x-grid points are still the same each process uses a different range of them. This might interfere with the caching of the EKOs, what do you think?

felixhekhorn:
caching? the best we do at the moment is just the slice here (and that is on pid so unaffected): https://github.com/N3PDF/pineappl/blob/d924e9e41798c1f755849fd054938c05bbdb99c2/pineappl/src/grid.rs#L1740
afterwards we have the index remap, which would need to be shifted down to subgrid level as far as I understand
https://github.com/N3PDF/pineappl/blob/d924e9e41798c1f755849fd054938c05bbdb99c2/pineappl/src/grid.rs#L1754

cschwan:
Sorry, I meant caching of the EKOs in pineko or probably somewhere higher. It might be a problem that every grid we'd like then has a different x-grid (well, technically a subset as I said).

AleCandido:
Of course we can do, if needed (that's why we implemented in the first place). But I'd try to avoid re-interpolating as much as possible...

cschwan:
> I.e.: everything fine, we'll do as @felixhekhorn said, if we have to. But since we agreed to have a single x-grid for the theory,

This won't change.

> it would be ideal to have all the PineAPPLgrid with the same one, if possible. This will automatically skip re-interpolation when not needed

That's the case, you should be able to feed the same set of EKOs to `Grid::convolute_eko`, and it'll select the necessary subset. But `Grid::eko_info` will return a smaller set of x-grid points, and I wonder whether higher up the toolchain this will lead to regeneration of many subsets of EKOs of the original 50x50 again and again for the newly implemented datasets with the implementation of pineko (?) right now. But it's certainly possible to avoid that.

[...]

PR #182, opened 2022-09-30
Fix bug in `Grid::convolute_eko`

AleCandido:
I reverted https://github.com/N3PDF/pineappl/commit/b7014fdc4ecf9ed3ef15f08cc47ee5761ed0b471, as @cschwan suspect this to be the issue https://github.com/N3PDF/pineappl/pull/103#discussion_r984360664.

### DIFF
--- a/pineappl/src/grid.rs
+++ b/pineappl/src/grid.rs
@@ -1531,46 +1531,6 @@ impl Grid {
             })
             .collect();
 
-        // the x1 and x2 grid values are the same across all non-zero subgrids
-        let first_non_zero_grid = self
-            .subgrids
-            .iter()
-            .find(|subgrid| !subgrid.is_empty())
-            .unwrap_or_else(|| unreachable!());
-        // source x1/x2 grid might differ and be differently sorted than the operator
-        let x1_grid = if has_pdf1 {
-            first_non_zero_grid
-                .x1_grid()
-                .iter()
-                .map(|x| {
-                    eko_info
-                        .grid_axes
-                        .x_grid
-                        .iter()
-                        .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
-                        .unwrap_or_else(|| unreachable!())
-                })
-                .collect()
-        } else {
-            Vec::new()
-        };
-        let x2_grid = if has_pdf2 {
-            first_non_zero_grid
-                .x2_grid()
-                .iter()
-                .map(|x| {
-                    eko_info
-                        .grid_axes
-                        .x_grid
-                        .iter()
-                        .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
-                        .unwrap_or_else(|| unreachable!())
-                })
-                .collect()
-        } else {
-            Vec::new()
-        };
-
         // iterate over all bins, which are mapped one-to-one from the target to the source grid
         for bin in 0..self.bin_info().bins() {
             // iterate over the source grid luminosities
@@ -1618,6 +1578,40 @@ impl Grid {
                     };
 
                     let src_subgrid = &self.subgrids[[order, bin, src_lumi]];
+
+                    // source x1/x2 grid might differ and be differently sorted than the operator
+                    let x1_grid = if has_pdf1 {
+                        src_subgrid
+                            .x1_grid()
+                            .iter()
+                            .map(|x| {
+                                eko_info
+                                    .grid_axes
+                                    .x_grid
+                                    .iter()
+                                    .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
+                                    .unwrap_or_else(|| unreachable!())
+                            })
+                            .collect()
+                    } else {
+                        Vec::new()
+                    };
+                    let x2_grid = if has_pdf2 {
+                        src_subgrid
+                            .x2_grid()
+                            .iter()
+                            .map(|x| {
+                                eko_info
+                                    .grid_axes
+                                    .x_grid
+                                    .iter()
+                                    .position(|xi| approx_eq!(f64, *xi, *x, ulps = 64))
+                                    .unwrap_or_else(|| unreachable!())
+                            })
+                            .collect()
+                    } else {
+                        Vec::new()
+                    };
 
                     for ((iq2, ix1, ix2), &value) in src_subgrid.iter() {
                         let scale = src_subgrid.mu2_grid()[iq2].fac;


AleCandido:
@felixhekhorn can you test on your example?

felixhekhorn:
With this branch I get
```
$ pineappl diff data/grids/208/ATLASWZRAP36PB-ATLAS-arXiv\:1109.5141-Z0_eta34.pineappl.lz4 data/fktables/208/ATLASWZRAP36PB-ATLAS-arXiv\:1109.5141-Z0_eta34.pineappl.lz4 NNPDF40_nlo_as_01180 --ignore-orders --ignore-lumis
LHAPDF 6.4.0 loading /home/felix/local/share/LHAPDF/NNPDF40_nlo_as_01180/NNPDF40_nlo_as_01180_0000.dat
NNPDF40_nlo_as_01180 PDF set, member #0, version 1; LHAPDF ID = 331700
b   x1                  diff              
-+---+---+-----------+-----------+--------
0   0 0.4 1.3463245e5 1.3450066e5 9.798e-4
1 0.4 0.8 1.3347254e5 1.3334425e5 9.621e-4
2 0.8 1.2 1.3077692e5 1.3066370e5 8.665e-4
3 1.2 1.6 1.2699118e5 1.2689062e5 7.925e-4
4 1.6   2 1.2116815e5 1.2109076e5 6.391e-4
5   2 2.4 1.1276176e5 1.1269892e5 5.575e-4
6 2.4 2.8 9.8484125e4 9.8421191e4 6.394e-4
7 2.8 3.6 5.7843820e4 5.7813500e4 5.244e-4
```
which seems to catch the problem - we're back at permille accuracy!

I suggest, @andreab1997, you can recompute (whenever you have time, of course) the problematic grids in https://github.com/NNPDF/fktables/issues/32 and confirm or deny the conclusion

I also can partly confirm another comment of @andreab1997: this took ~2h to compute which he observed for the good commits as well (in contrast to "short runs" which resulted in wrong FK tables)

andreab1997:
> With this branch I get
> 
> ```
> $ pineappl diff data/grids/208/ATLASWZRAP36PB-ATLAS-arXiv\:1109.5141-Z0_eta34.pineappl.lz4 data/fktables/208/ATLASWZRAP36PB-ATLAS-arXiv\:1109.5141-Z0_eta34.pineappl.lz4 NNPDF40_nlo_as_01180 --ignore-orders --ignore-lumis
> LHAPDF 6.4.0 loading /home/felix/local/share/LHAPDF/NNPDF40_nlo_as_01180/NNPDF40_nlo_as_01180_0000.dat
> NNPDF40_nlo_as_01180 PDF set, member #0, version 1; LHAPDF ID = 331700
> b   x1                  diff              
> -+---+---+-----------+-----------+--------
> 0   0 0.4 1.3463245e5 1.3450066e5 9.798e-4
> 1 0.4 0.8 1.3347254e5 1.3334425e5 9.621e-4
> 2 0.8 1.2 1.3077692e5 1.3066370e5 8.665e-4
> 3 1.2 1.6 1.2699118e5 1.2689062e5 7.925e-4
> 4 1.6   2 1.2116815e5 1.2109076e5 6.391e-4
> 5   2 2.4 1.1276176e5 1.1269892e5 5.575e-4
> 6 2.4 2.8 9.8484125e4 9.8421191e4 6.394e-4
> 7 2.8 3.6 5.7843820e4 5.7813500e4 5.244e-4
> ```
> 
> which seems to catch the problem - we're back at permille accuracy!
> 
> I suggest, @andreab1997, you can recompute (whenever you have time, of course) the problematic grids in [NNPDF/fktables#32](https://github.com/NNPDF/fktables/issues/32) and confirm or deny the conclusion
> 
> I also can partly confirm another comment of @andreab1997: this took ~2h to compute which he observed for the good commits as well (in contrast to "short runs" which resulted in wrong FK tables)

Ok very good, thank you very much for this. As soon as I have time I will run a lot of tests to really be sure, but I would say that most likely we found the issue.

AleCandido:
> I also can partly confirm another comment of @andreab1997: this took ~2h to compute which he observed for the good commits as well (in contrast to "short runs" which resulted in wrong FK tables)

@cschwan can we maybe consider an optimization for the grid, in which we reinterpolate all the subgrids on a common xgrid?

For sure this will bring some loss in accuracy, but might be an acceptable one (against a considerable speed up).

cschwan:
@andreab1997 @felixhekhorn as a guiding principle: all grids converted from APPLgrids are potentially affected by this bug, grids generated from `pinefarm` (with yadism or Madgraph5@MC_NLO) *should* be OK.

cschwan:
> > I also can partly confirm another comment of @andreab1997: this took ~2h to compute which he observed for the good commits as well (in contrast to "short runs" which resulted in wrong FK tables)
> 
> @cschwan can we maybe consider an optimization for the grid, in which we reinterpolate all the subgrids on a common xgrid?
> 
> For sure this will bring some loss in accuracy, but might be an acceptable one (against a considerable speed-up).

As far as I understand this won't help (yet), `Grid::convolute_eko` is simply inherently slow. However, I've written a realisitic unit test and with that I'll be able to improve the performance. Monday is a holiday in Germany and by then I hopefully have something faster!

AleCandido:
> As far as I understand this won't help (yet), `Grid::convolute_eko` is simply inherently slow.

The idea would be to trash this PR and rely on the assumptions that they are all the same. According to @felixhekhorn 

> this took ~2h to compute which he observed for the good commits as well (in contrast to "short runs" which resulted in wrong FK tables)

so this bug fix is heavily affecting performances. If you can live without, so much the better.

cschwan:
> > As far as I understand this won't help (yet), `Grid::convolute_eko` is simply inherently slow.
> 
> The idea would be to trash this PR and rely on the assumptions that they are all the same. According to @felixhekhorn

This I don't understand: why wouldn't we want to merge this pull request?

> 
> > this took ~2h to compute which he observed for the good commits as well (in contrast to "short runs" which resulted in wrong FK tables)
> 
> so this bug fix is heavily affecting performances. If we can live without, so much the better.

I can always write a program that calculates the wrong results very quickly, but that isn't very helpful, of course.

felixhekhorn:
> > As far as I understand this won't help (yet), `Grid::convolute_eko` is simply inherently slow.
> 
> The idea would be to trash this PR and rely on the assumptions that they are all the same. According to @felixhekhorn

hm? if they are coming from the outside (meaning APPL) there is not much we can do, can we? reinterpolate is an option, but not the short term answer

andreab1997:
> With this branch I get
> 
> ```
> $ pineappl diff data/grids/208/ATLASWZRAP36PB-ATLAS-arXiv\:1109.5141-Z0_eta34.pineappl.lz4 data/fktables/208/ATLASWZRAP36PB-ATLAS-arXiv\:1109.5141-Z0_eta34.pineappl.lz4 NNPDF40_nlo_as_01180 --ignore-orders --ignore-lumis
> LHAPDF 6.4.0 loading /home/felix/local/share/LHAPDF/NNPDF40_nlo_as_01180/NNPDF40_nlo_as_01180_0000.dat
> NNPDF40_nlo_as_01180 PDF set, member #0, version 1; LHAPDF ID = 331700
> b   x1                  diff              
> -+---+---+-----------+-----------+--------
> 0   0 0.4 1.3463245e5 1.3450066e5 9.798e-4
> 1 0.4 0.8 1.3347254e5 1.3334425e5 9.621e-4
> 2 0.8 1.2 1.3077692e5 1.3066370e5 8.665e-4
> 3 1.2 1.6 1.2699118e5 1.2689062e5 7.925e-4
> 4 1.6   2 1.2116815e5 1.2109076e5 6.391e-4
> 5   2 2.4 1.1276176e5 1.1269892e5 5.575e-4
> 6 2.4 2.8 9.8484125e4 9.8421191e4 6.394e-4
> 7 2.8 3.6 5.7843820e4 5.7813500e4 5.244e-4
> ```
> 
> which seems to catch the problem - we're back at permille accuracy!
> 
> I suggest, @andreab1997, you can recompute (whenever you have time, of course) the problematic grids in [NNPDF/fktables#32](https://github.com/NNPDF/fktables/issues/32) and confirm or deny the conclusion
> 
> I also can partly confirm another comment of @andreab1997: this took ~2h to compute which he observed for the good commits as well (in contrast to "short runs" which resulted in wrong FK tables)

I am computing now all the fktables that are listed in "Things to check" in https://github.com/NNPDF/fktables/issues/32 with this branch and I am ticking its elements one by one as I check them. For the moment the three that I have computed are fine so I believe also the others will be fine as well. Thank you again!

AleCandido:
> This I don't understand: why wouldn't we want to merge this pull request?
> I can always write a program that calculates the wrong results very quickly, but that isn't very helpful, of course.

Of course, we want correct results, but we want also to be fast. That's why I proposed reinterpolation: if we can transform subgrids, such that they all have a common x, we will loose a bit in precision, but for a great gain in performances.

Zaharid:
fwiw the more I think about it the more I am convinced that what we want at the fit level is a common x grid. The fitting problem becomes a lot easier to reason about then.

Measuring how fine the grid needs to be is relatively easy.

AleCandido:
> fwiw the more I think about it the more I am convinced that what we want at the fit level is a common x grid. The fitting problem becomes a lot easier to reason about then.

I believe we all agree, that is definitely a good thing. But it is about the xgrid in the FkTable, that is coming from the EKO output, and it is already guaranteed.

Here we are talking of the EKO input (`eko` is already reinterpolating, so allows for input xgrid different from output), that has to be matched with the equivalent of an APPLgrid.
Unfortunately, it seems that not all the old APPLgrids where using a single xgrid for all the subgrids...

By the way @cschwan, I believe that the current status might also be wrong: if subgrid A has a different xgrid than subgrid B, in order to use a single EKO we have to compute it on the union (since we are not reinterpolating it on the fly).
But then the subgrid is convoluted only overr a subset of the EKO, and this should be wrong: even if you have a subset of points in the extended one, the interpolation basis is not a subset of the basis, because the support of the polynomials in the smaller basis has to be larger than for the corresponding in the larger one.
Thus I believe we **need** reinterpolation, and also what we were doing before might have been wrong (and now I doubt also what was doing APFELcomb on those grids...)

cschwan:
@AleCandido I think what you're probably doing is to calculate the EKO for each bin (they don't change over luminosities and there's only one order) and then merge all EKOs. I think in this case there shouldn't be a problem, since most likely the x grids for each bin are all pairwise disjoint sets. If they aren't the merging of the EKOs for each bin should be going wrong.

AleCandido:
> @AleCandido I think what you're probably doing is to calculate the EKO for each bin (the x grids don't change over luminosities and there's only one order) and then merge all EKOs. I think in this case there shouldn't be a problem, since most likely the x grids for each bin are all pairwise disjoint sets. If they aren't the merging of the EKOs for each bin should be going wrong.

Don't think so, since we rely on grid information to construct the EKO, in particular on `GridAxes`, that
- do depend only on the grid:
  https://github.com/N3PDF/pineappl/blob/ad22780918c4452cca0c4b6ddded6a600237f143/pineappl/src/grid.rs#L1303
- contains a single `xgrid`:
  https://github.com/N3PDF/pineappl/blob/ad22780918c4452cca0c4b6ddded6a600237f143/pineappl/src/grid.rs#L355-L357

So, given a grid, PineAPPL is always returning us a single xgrid, not a bin-dependent one.
The only thing we can do at this point, it is to compute one single EKO for the single xgrid given.

felixhekhorn:
> Thus I believe we **need** reinterpolation, and also what we were doing before might have been wrong (and now I doubt also what was doing APFELcomb on those grids...)

I don't believe this is a problem, because we're already at the level of matrix multiplication (after the grid), so the interpolation only plays a role inside the grid. Afterwards, it is just linear algebra: meaning eko is exact on the grid points (by construction) and so the information from the input to any output point is exact

cschwan:
Let's say you have a grid with two bins for DIS. The first bin has this x grid:

    [0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8]

and the second bin:

    [0.25, 0.35, 0.45, 0.55, 0.65, 0.75, 0.85]

which means you calculate the EKO for the union of the x grid points given above. For the first bin that means the evolution is done using the a subset (the first set), while the other x grid points contribute with zero. I don't see anything wrong with that; the interpolation should still work.

AleCandido:
> I don't believe this is a problem, because we're already at the level of matrix multiplication (after the grid), so the interpolation only plays a role inside the grid. Afterwards, it is just linear algebra: meaning eko is exact on the grid points (by construction) and so the information from the input to any output point is exact

One of the two is exact, this is true: matrix multiplication is always a contraction over two paired, but different, indices:
1. on the interpolation grid (function evaluate at that point) and
2. on the interpolation basis (coefficient of that polynomial)

This only works if the index over the EKO is of kind 1., can you convince me of this? @felixhekhorn

AleCandido:
> you calculate the EKO for the union of the x grid points given above. For the first bin that means the evolution is done using the a subset (the first set), while the other x grid points contribute with zero

@cschwan the problem is there for one of the EKO indices (input or output, it is always confusing to me).
On one of the two, the EKO is the coefficient over one basis of functions: if you compute the coefficients of a given function over an extended basis, but then you only use a subset, you are effectively trashing the others, so you are telling that they should be multiplied by zeros. But that is not true: if the yadism function is evaluated at a limited subset of points, it doesn't mean that is zero elsewhere.

cschwan:
@AleCandido if we use `data/grids/208/ATLASWZRAP36PB-ATLAS-arXiv:1109.5141-Z0_eta34.pineappl.lz4`, delete everything but, say, the first bin, and then evolve the grid with one bin using

1. the already computed EKO and compare that against
2. the result obtained using a newly-computed EKO *only* for the first bin

we should be able to see what happens and how big the difference is.

AleCandido:
> we should be able to see what happens and how big the difference is.

Let me think one moment, and let's wait for @felixhekhorn reply.

In any case, I don't believe the exercise to be extremely relevant: either it is correct, or it is wrong.
Empirically correct is telling us little, because I know that if it is wrong I can make the result arbitrarily different, so we would rely on the specific xgrids chosen (and they are not chosen by us).
Moreover, empirically correct definition is based on agreement with APFELcomb, but we may well have perfect agreement, and just being both wrong in the same way.

cschwan:
@AleCandido we had a similar discussion here: https://github.com/N3PDF/pineappl/issues/151.

AleCandido:
> @AleCandido we had a similar discussion here: #151.

In that case, we were considering different grids. Now the problem is within a single grid.
But I agree it is a similar problem.

cschwan:
> In any case, I don't believe the exercise to be extremely relevant: either it is correct, or it is wrong. Empirically correct is telling us little, because I know that if it is wrong I can make the result arbitrarily different, so we would rely on the specific xgrids chosen (and they are not chosen by us). Moreover, empirically correct definition is based on agreement with APFELcomb, but we may well have perfect agreement, and just being both wrong in the same way.

If in all cases we've seen so far the differences are small then I believe - which is a rather weak statement, I admit - the interpolation can't be too far off and at the very least we probably don't have to hold off producing fits with the new pipeline.

andreab1997:
Since we are now pretty sure that this was the source of the bug, @cschwan are you going to release a new version? Or do you want to wait to improve performances first? For me it makes essentially no difference (although it is obviuosly simpler with a release).

cschwan:
@andreab1997 I'll probably release a new version at the end of today.

@AleCandido @felixhekhorn given that we fixed the initial problem, are you OK with merging this pull request?

AleCandido:
> @AleCandido @felixhekhorn given that we fixed the initial problem, are you OK with merging this pull request?

Yes, let's merge this one, and go back to the former status. We can keep discussing about the rest separately.

cschwan (approved):

cschwan:
@andreab1997 I've just released `v0.5.6` which includes this fix. Crates and packages should become available soon!

andreab1997:
> @andreab1997 I've just released `v0.5.6` which includes this fix. Crates and packages should become available soon!

Ok thank you very much!

PR #187, opened 2022-10-26
Try using `rust-cov` instead of `grcov`

cschwan:
This should improve the precision of the code coverage.


AleCandido:
Why did you remove APPL and fastNLO conversion from the workflow?

cschwan:
> Why did you remove APPL and fastNLO conversion from the workflow?

Only to speed up the CI - they aren't relevant for this branch. I will add them back before I merge this branch into master.

Issue #118, opened 2022-02-18
Release version v1 of the file format


cschwan:
Here's what I'd like to change in a newer version of the *file format*, `v1`, replacing the current (`v0`) one:

- [ ] merge the types `BinInfo` and `BinRemapper` into `BinLimits`. The reason for having them separate is historical only (see https://github.com/N3PDF/pineappl/issues/83#issue-1039614822) and merging them will make parts of the code much easier
- [x] merge `Mmv3` into `Grid`, which means that metadata will always be present making metadata-related code much shorter
- [x] change the `Order`'s member from `u32` to `u8` and add another member to support #98
- [x] remove most of the `Subgrid` types and keep only the ones we use (see https://github.com/N3PDF/pineappl/issues/83#issuecomment-1007954917)
- [x] ~merge all `Subgrid` types into a single module~
- [x] ~export most important types into the main module~
- [ ] ~use bincode version 2 with [varint encoding](https://docs.rs/bincode/latest/bincode/config/struct.VarintEncoding.html), see also item here: https://github.com/N3PDF/pineappl/issues/45#issue-724493266~
- [x] make some of the metadata proper members of `Grid`: `initial_state_1`, `initial_state_2` and possibly `lumi_id_types`, which are important in some places; https://github.com/N3PDF/pineappl/issues/135 is a related Issue; done in #299
- [x] replace `set_convolutions` with `convolutions_mut`; done in #299
- [x] Unify `SubgridParams` with `ExtraSubgridParams` and rename them to `InterpolationParams`; make their parameters `pub` instead of having many setters and getters; this has now been merged into single `Interp` type
- [x] Generalize convolutions to support fragmentation functions (see https://github.com/N3PDF/pineappl/issues/172)
- [x] Rename `LumiCache` to `ConvolutionCache`, which better reflects its purpose
- [x] generalize `Channel` to allow more or less than 2 PIDs per linear combination; done in #299
- [x] Remove `Convolutions::None`
- [ ] <https://github.com/NNPDF/pineappl/issues/270>

_Originally posted by @cschwan in https://github.com/N3PDF/pineappl/issues/83#issuecomment-1007949008_

cschwan:
Technically we can have different file formats at the same time by versioning also the modules:
- `pineappl::v0::` would contain the old code
- `pineappl::v1::` would contain the new file format
- and so on should we need even more changes.
- in the module `pineappl` we can insert `pub use ...` statements to import the most recent file format.

[...]

Issue #97, opened 2022-01-04
Fix `Grid::convolute_subgrid`


cschwan:
The method `Grid::convolute_subgrid` returns wrong results for many subgrid types because reweighting isn't considered, see also the following comment:

https://github.com/N3PDF/pineappl/blob/7b8b306c008ffded9efcd62cb43b0c9b06eb7f6d/pineappl/src/grid.rs#L636

This affects the following subgrid types for which it must be fixed:

- [ ] `LagrangeSubgridV1`
- [ ] `LagrangeSubgridV2`
- [ ] `LagrangeSparseSubgridV1`

This doesn't affect `ImportOnlySubgridV1` and `ImportOnlySubgridV2`, however, so that while technically this is a bug, all optimized grids aren't affected by it.

PR #194, opened 2022-12-07
Fix `Lagrange*` subgrid iterators

cschwan:
This fixes <https://github.com/NNPDF/pineappl/issues/97>, but more importantly it will enable importing grids using `indexed_iter`. This will be needed for the new file format, see <https://github.com/NNPDF/pineappl/issues/118>.


cschwan:
Another bug fix was added in commit ace8d9d49aa341c0aac3025dc24d4afa4db1be4b.

PR #196, opened 2022-12-16
Avoid copies in Python interface

cschwan:
This PR changes the Rust interface to use `ArrayView5` instead of `Array5`, which has the advantage of avoiding a full copy of the operator in `PyGrid::evolve`.

### DIFF
--- a/pineappl/src/evolution.rs
+++ b/pineappl/src/evolution.rs
@@ -8,7 +8,7 @@ use super::sparse_array3::SparseArray3;
 use super::subgrid::{Mu2, Subgrid, SubgridEnum};
 use float_cmp::approx_eq;
 use itertools::Itertools;
-use ndarray::{s, Array1, Array2, Array3, Array5, ArrayView1, Axis};
+use ndarray::{s, Array1, Array2, Array3, ArrayView1, ArrayView5, Axis};
 use std::iter;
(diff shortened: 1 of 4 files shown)


AleCandido:
I agree it would be desirable, and, for FkTable in general, you want only to read them. But are you sure it is supported by `rust-numpy`?

AleCandido:
I.e. are you sure that `rust-numpy` won't generate an actual Python object with its own memory anyhow?

cschwan:
I'm certain that in `PyGrid` the methods `into_owned()` and `to_vec()` make copies, and we can easily avoid them.

AleCandido:
Oh yes, I was reading right now the code, before I stopped to the comment. Essentially, it is the "opposite" problem of what I thought: you really have Python arrays, and don't want the copy to be done in Rust.

It might be that is also the third copy, if a copy is already done when passing the object from Python down to Rust, I'm not sure of this, but for sure you can avoid at least this (and I would expect to be the only one, but we can always profile it).

However, I started reading `rust-numpy` docs in the meanwhile:
https://docs.rs/numpy/latest/numpy/borrow/index.html
and we are already using `PyReadonlyArray`, so this supports my hypothesis that there is no further copy.

PR #208, opened 2023-02-18
Attempt to use a custom container for CI

cschwan:
Use a custom container to cut down the time of the Rust and CAPI action.


cschwan:
- CAPI: https://github.com/NNPDF/pineappl/actions/runs/4210896135 vs. https://github.com/NNPDF/pineappl/actions/runs/4133228405: from 10 minutes to just 3 minutes, and
- Rust https://github.com/NNPDF/pineappl/actions/runs/4210896252 vs. https://github.com/NNPDF/pineappl/actions/runs/4133228418: from 18 minutes to 12 minutes.

Pretty good!

AleCandido:
In case you care, all the support for the CI of the NNPDF projects is currently collected in the repo https://github.com/NNPDF/workflows (other than NNPDF/nnpdf itself, that always had a separate management, people-wise and therefore tools-wise).

In particular, the `Containerfile`s for the containers we are using are all in 
https://github.com/NNPDF/workflows/tree/v2/containers

The containers themselves belong to the NNPDF org, and they are linked to the workflows repo:
https://github.com/orgs/NNPDF/packages?repo_name=workflows

cschwan:
> In case you care, all the support for the CI of the NNPDF projects is currently collected in the repo https://github.com/NNPDF/workflows (other than NNPDF/nnpdf itself, that always had a separate management, people-wise and therefore tools-wise).
> 
> In particular, the `Containerfile`s for the containers we are using are all in https://github.com/NNPDF/workflows/tree/v2/containers
> 
> The containers themselves belong to the NNPDF org, and they are linked to the workflows repo: https://github.com/orgs/NNPDF/packages?repo_name=workflows
> 
> Btw, there is already one container with just LHAPDF installed, together with the baseline of common PDF sets. But there is no installation of Rust available, so keep using your own.

In that case I think it makes sense to migrate it there!

AleCandido:
> In that case I think it makes sense to migrate it there!

To keep things as plain as possible, I'd suggest completing it here and merge, then migrate in a separate PR.

In any case, thanks a lot for this, this will be useful also for other possible uses, not only the CI!

cschwan:
Something I don't completely understand yet is which files github needs and therefore which files in the container I can delete. I understand that when I chroot into the container I only have the files in the container, obviously, but I think github doesn't exactly do that and needs fewer files. The fact that I still have to specify `runs-on: ubuntu-latest` probably means that some files of that system are available...

AleCandido:
You're definitely looking at the process from a too low-level perspective. Usually you just talk of container runtimes, and forget about the `chroot` part (even though is definitely useful to remember with in general, if you are concerned about privileges escalation and security).

You do not need any file specifically in the container, it can also be almost empty, for as long as it can run a shell (I believe that's the only program executed).

The reason why you *have to* specify `runs on:` is that they have to allocate a virtual machine, and you are billed according to the virtual machine that you ask for.
Moreover, I'm not sure whether they are supported, but in principle you might also want to run a Windows container, and for that you need a Windows virtual machine.

maintainer/pineappl-ci/Containerfile
AleCandido, line 1: We could convert the large `RUN` to a shell script :)

(moreover, I'm still convinced NNPDF/workflows is a good place for a lot of this, but I'm not insisting, because for the time being it might be more convenient to have everything in a single PR)

cschwan: I was about to ask you, I'm no longer sure about that. On the hand it's nice to have all containers created in a central places, on the other hand this container is very much tailored to be used for PineAPPL's CI. As far as I understand, <https://github.com/NNPDF/workflows/> is being used across many repositories.

Are there are points I'm missing?

AleCandido: The container might be nice to be used potentially in other projects as well, e.g. those who depend on PineAPPL, but it's not my main concern.

I'm much more interested in splitting the "packages" (install scripts) and store aside the other ones we already have https://github.com/NNPDF/workflows/tree/v2/packages

Those install scripts are pretty small and simple, but they give a unified way to install software we all use, potentially also in a virtual environment (with the env var `PREFIX`, take LHAPDF as an example). So, they are flexible tools for many scenarios.

Then, once I have all *packages* there, it has been natural for me to define also the containers there, but if you prefer to keep it here I have no objection.
I'd just keep in a single place, to avoid inconsistency. And the same for the install scripts (but in that case I'm opinionated in favor of NNPDF/workflows).

cschwan: OK, let's put it into NNPDF/workflows. Could you take care of it?

AleCandido: For sure. But are the install scripts final?

cschwan: The Containerfile should be final, but every time we need to increase the MSRV or one of the sofware versions/PDF sets we need to update the container, of course.

cschwan: In commit aab8f5fb0e4d295689fe1d7bc281f1cc6af79b91 I moved the most of the container instructions into a bash script, which is now much more readable.

cschwan: I will close this discussion. Moving the container building instructions must be done in <https://github.com/NNPDF/workflows>.

AleCandido: Thanks, that's indeed helpful.

In principle, it makes little difference having the script inside the Containerfile, or an extremely simple Containerfile invoking the script, but it becomes extremely useful if the Containerfile grows with specific instructions (even though the main reason for which I like it is to have proper syntax highlighting and my Bash formatters running automatically...)

cschwan: ... it also allows to place comments inside the script, which is helpful to remind us why we need `rustup component add llvm-tools-preview`, for instance!

PR #212, opened 2023-03-01
Move wheels release to maturin action

AleCandido:


AleCandido:
The `maturin` generator definitely added to many architectures to the workflow. I believe we don't care about all of them.

@cschwan @scarlehoff can you have a look and comment if we only care about `x86_64` (and equivalents) or we also want to retain some of others?
Concerning MacOS, I decided to release universal wheels (don't care too much if it is suboptimal or what on M1/M2, for as long as it works).

AleCandido:
This is the version we are using https://github.com/PyO3/maturin-action/releases/tag/v1

cschwan:
The version is indeed fine, but it seems that `working-directory` is missing. I found this Issue :smile:: <https://github.com/PyO3/maturin-action/issues/11>.

cschwan:
`--manifest-path` is set in `$FLAGS` as an environment variable, but is it possible that it isn't properly substituted? The log says:

    Caused by: Failed to parse Cargo.toml at /home/runner/work/pineappl/pineappl/Cargo.toml

while it should be the `pineappl_py` subfolder. Assuming that `/home/runner/work/pineappl/` is the repository root I don't understand why it goes into the Rust crate subfolder.

cschwan:
It seems that `$FLAGS` wasn't expanded, the last commit doesn't use it and work well (so far).

AleCandido:
Ok, fine for `$FLAGS`, I will find a better way to do it, but we definitely can't use `--find-interpreter` on Linux, because of 

> Note that the actions/setup-python action won't affect manylinux build since it's containerized, so if you want to build for certain Python version for Linux, use -i pythonX.Y in the args option in PyO3/maturin-action instead, for example

https://github.com/PyO3/maturin-action#manylinux-docker-container

For the other two, we can just add Python versions to the matrix.

cschwan:
On Linux I see the following messages:
```
📦 Built wheel for CPython 3.7m to dist/pineappl-0.6.0a1-cp37-cp37m-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
📦 Built wheel for CPython 3.8 to dist/pineappl-0.6.0a1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
📦 Built wheel for CPython 3.9 to dist/pineappl-0.6.0a1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
📦 Built wheel for CPython 3.10 to dist/pineappl-0.6.0a1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
📦 Built wheel for CPython 3.11 to dist/pineappl-0.6.0a1-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
📦 Built wheel for PyPy 3.7 to dist/pineappl-0.6.0a1-pp37-pypy37_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
📦 Built wheel for PyPy 3.8 to dist/pineappl-0.6.0a1-pp38-pypy38_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
📦 Built wheel for PyPy 3.9 to dist/pineappl-0.6.0a1-pp39-pypy39_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

which should be more than we need, should it not?

The documentation of PyO3 that you quoted I read as: if we wanted *one specific version* of Python, say *only* 3.11, then we'd have to use option. Or do I misread it?

.github/workflows/wheels.yml
```diff
@@ -1,47 +1,89 @@
...
 jobs:
-  deploy-linux:
+  linux:
     runs-on: ubuntu-latest
-    # a dedicated container is needed, so maturin-action is not an option, see pineappl_py/package/README.md for details
-    container:
-      image: docker://ghcr.io/n3pdf/maturin:0.2.1
+    strategy:
+      matrix:
+        target: [x86_64]
```
AleCandido, line 16: If you just need a single value, you can also inline this in the place where it is used.
Or do you believe we might need more architectures in the future, and thus it makes sense to keep as a list?

.github/workflows/wheels.yml
```diff
@@ -1,47 +1,89 @@
...
-        env:
-          HOME: /home/pineappl
-        run: |
-          ~/maturin publish --skip-existing --username __token__ --password ${{ secrets.PYPI_TOKEN }} --interpreter 3.7 3.8 3.9 3.10 pypy3.7
+      - uses: actions/checkout@v3
+      - name: Build wheels
+        uses: PyO3/maturin-action@v1
+        with:
+          target: ${{ matrix.target }}
+          args: ${{ env.flags }} --interpreter 3.7 3.8 3.9 3.10 3.11 pypy3.8
```
AleCandido, line 23: The `interpreter` arguments are a few, can't we define a variable for them? Don't you believe they could be shared across different platforms? (at least the CPython versions, exclude `pypy`, for which one wheel is definitely enough, since we do not have any evidence for anyone using it)

AleCandido: In particular, if you want to have a list, and not a single value, you can use `env` instead of `matrix`, as it was done in 03616cc8bfe8993b28b85a876f479fdfa38c007b

AleCandido: Since by now you're not using `matrix` in any job, you won't need all the `fromJson` alchemy, but just define the string as you're using it (the exact same characters) and use the `env` context to expand:
```
${{ env.python_versions }}
```

Since contexts are expanded by the workflow runner, and not by the shell, there is no doubt it will be expanded before execution.

.github/workflows/wheels.yml
cschwan: Exactly, we might want to support more architectures at some point!

.github/workflows/wheels.yml
cschwan: I wanted to see whether passing multiple versions is possible at all (it is!).

cschwan: After our (private) conversation we decide to go back to `--find-interpreter` which selects all available Python versions.

cschwan:
@AleCandido I think this finally ready to merge. I'd like to squash and merge it since most changes are the typical trial-and-error Github action commits. If you're OK with it, I'd merge as soon as the CI has finished.

AleCandido:
I'm fine with squashing, I also dislike trial-and-errors commits required to debug workflows.
I still hope that one day they will either give a debug shell or officially support `act`.

Give me just the time to have a final look, and then we will merge.

AleCandido (commented):
I don't have the approve button, because I'm the PR author, but... I approve!

Issue #211, opened 2023-02-28
Problem with Python deployment


cschwan:
The Python deployment currently doesn't work, because maturin is too old. We need at least `0.13.4`, see also commit fc50e9c5d785076968f52ad9b913e32c0415bfa5. @AleCandido, could you take care of this? Podman doesn't work reliably on my system at work and I'm not sure I can upload it.

cschwan:
@andreab1997 this is probably causing <https://github.com/NNPDF/pineko/actions/runs/4292741468/jobs/7479555051> to fail.

AleCandido:
> @andreab1997 this is probably causing [NNPDF/pineko/actions/runs/4292741468/jobs/7479555051](https://github.com/NNPDF/pineko/actions/runs/4292741468/jobs/7479555051) to fail.

There I believe there might be two problems:
- indeed, wheels for Linux have not been deployed https://pypi.org/project/pineappl/0.6.0a1/#files
- pre-releases have not been allowed in pyproject.toml for PineAPPL, but I don't know if with the exact number this option is not needed (however, we should always use `^x.y.z` rather than `x.y.z`)

I will update the container (but this is a rather simple task, just have a look here https://github.com/NNPDF/pineappl/tree/master/pineappl_py/package)

If possible, I'd also like to get rid of that container at some point, but at the moment the quickest option is just to update it...

cschwan:
> If possible, I'd also like to get rid of that container at some point, but at the moment the quickest option is just to update it...

I was thinking of the same. We can use #208 for that, by basing the container on a manylinux image instead of the Rust image.

AleCandido:
> We can use #208 for that, by basing the container on a manylinux image instead of the Rust image.

That is exactly the `maturin` container: it is just based on `manylinux` with Rust and Maturin installed.

Instead, I was thinking to revisit the [`maturin-action`](https://github.com/NNPDF/pineappl/blob/master/pineappl_py/package/Containerfile), most likely we can fully get rid of the whole container at all, and just delegate to it (it is also generate the workflow by itself).

andreab1997:
Ok then let me know what I should do for [NNPDF/pineko/actions/runs/4292741468/jobs/7479555051](https://github.com/NNPDF/pineko/actions/runs/4292741468/jobs/7479555051)

AleCandido:
@andreab1997 For what it concerns you, there has been no release yet. Unless you want to run tests on Windows.

cschwan:
[...]

PR #214, opened 2023-03-06
Fix python version problems

cschwan:
Here we'll fix the problems discovered in this comment: <https://github.com/NNPDF/pineappl/issues/211#issuecomment-1453282553>.


.github/workflows/wheels.yml
```diff
@@ -13,7 +11,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: ['3.7', '3.8', '3.9', '3.11', 'pypy3.7', 'pypy3.8', 'pypy3.9', '3.10']
```
AleCandido: This should be irrelevant, since the wheel is built in any case inside a `manylinux` container.

Not sure how to implement it with `sed`...
I know something algorithmic is more convenient (even with some changes to the base, `sed` would keep working), but if it is too complex I'd rather go for a patch, or just commit the final result manually edited.

.github/workflows/wheels.yml
```diff
@@ -69,19 +67,19 @@ jobs:
...
-        uses: PyO3/maturin-action@v1
-        env:
-          MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
-        with:
-          command: upload
-          args: --skip-existing *
+  #release:
+  #  name: Release
+  #  runs-on: ubuntu-latest
+  #  if: "startsWith(github.ref, 'refs/tags/')"
```
AleCandido, line 97: Since this is only running on releases, you could keep this uncommented (just to reduce the diff)

.github/workflows/wheels.yml
```diff
@@ -1,7 +1,5 @@
 # Generated by maturin v0.14.14 and modified by update-wheels-ci.sh
-on:
-  release:
-    types: [published]
+on: push
```
AleCandido, line 2: You can also add an event, without removing the previous one>
```suggestion
on:
  push:
  release:
    types: [published]
```

.github/workflows/wheels.yml
cschwan: About the irrelevancy I'm not so sure anymore, it seems `setup-python` may also have an effect there: <https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-setup-python-with-a-self-hosted-runner>. We should figure out the effect and do it properly as you said.

.github/workflows/wheels.yml
cschwan: That's going in the right direction; I think I want to run every job except `release` in `wheels.yml` whenever I want to test them (in branches like this one).

.github/workflows/wheels.yml
AleCandido: A *self-hosted runner* is not a container, but a different machine you provide on your own to run the workflow, instead of using those of GitHub (i.e. Azure...).

Instead, `maturin-action` is running a container in a GitHub's virtual machine, that is relevant for the host system (we need a Linux machine to run a Linux container, otherwise we would need a Linux virtual machine inside the virtual machine), but it is not relevant Python-wise, since the `manylinux` container comes pre-built with its versions of Python.

This is true only for Linux, while for Windows and MacOS the `setup-python` is a crucial step.

.github/workflows/wheels.yml
AleCandido: This is also what Stefano C is doing in quantum repos (even if they are just pure Python... here I believe it's more relevant).

You can filter the jobs and individual steps with `if:` key, as you are doing, and the condition can contain an expression, that can depend on variables in contexts (thus including events and outputs of previous steps).
https://docs.github.com/en/actions/learn-github-actions/expressions
https://docs.github.com/en/actions/learn-github-actions/contexts
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

cschwan:
It seems to work, except that on Windows no PyPy (3.7, 3.8, 3.9) wheels are created.

AleCandido:
> It seems to work, except that on Windows no PyPy (3.7, 3.8, 3.9) wheels are created.

That is not a problem :)

AleCandido:
I still don't see MacOS wheels for CPython 3.7, 3.8, and 3.9 in https://github.com/NNPDF/pineappl/suites/11369716006/artifacts/584873397

cschwan:
You're right: instead the corresponding PyPy wheels are created (and also found) twice. See this line from the action:

    Found PyPy 3.7 at /Users/runner/hostedtoolcache/PyPy/3.7.13/x64/bin/python3.7, PyPy 3.8 at /Users/runner/hostedtoolcache/PyPy/3.8.16/x64/bin/python3.8, PyPy 3.9 at /Users/runner/hostedtoolcache/PyPy/3.9.16/x64/bin/python3.9, CPython 3.10 at /Users/runner/hostedtoolcache/Python/3.10.10/x64/bin/python3.10, CPython 3.11 at /Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11, PyPy 3.7 at /Users/runner/hostedtoolcache/PyPy/3.7.13/x64/bin/pypy3.7, PyPy 3.8 at /Users/runner/hostedtoolcache/PyPy/3.8.16/x64/bin/pypy3.8, PyPy 3.9 at /Users/runner/hostedtoolcache/PyPy/3.9.16/x64/bin/pypy3.9

Note how it finds PyPy (?) 3.7
- once as `/Users/runner/hostedtoolcache/PyPy/3.7.13/x64/bin/python3.7` and
- then as `/Users/runner/hostedtoolcache/PyPy/3.7.13/x64/bin/pypy3.7`.

Both claim to have

    Built wheel for PyPy 3.7 to dist/pineappl-0.6.0a2-pp37-pypy37_pp73-macosx_10_9_x86_64.macosx_11_0_arm64.macosx_10_9_universal2.whl

cschwan:
> > Note how it finds PyPy (?) 3.7
> > 
> > * once as `/Users/runner/hostedtoolcache/PyPy/3.7.13/x64/bin/python3.7` and
> > * then as `/Users/runner/hostedtoolcache/PyPy/3.7.13/x64/bin/pypy3.7`.
> 
> This I believe to be normal, and most likely the first is a link to the second, in such a way that PyPy can work both side-by-side and as a drop-in replacement of CPython.

Is this OK? Because the wheel is twice created for PyPy, can CPython work with it?

AleCandido:
> Is this OK? Because the wheel is twice created for PyPy, can CPython work with it?

No, definitely not. The compiled wheel needs ABI compatibility with the interpreter.
In order to guarantee this, wheels adopted a consistent naming scheme, that is the one you observed in the generated objects.

It might be that on MacOS, for whatever reason, PyPy is actually the `python3.x   x \in [7, 8, 9]` executable found by `maturin`, so it is regenerating twice the wheel for PyPy and overwriting it, while those for CPython are never generated.

Since for PyPy Linux is definitely far more than enough, maybe just try to remove it from all the lists of `python-version` (for Linux it will be built anyhow, since PyPy will be inside the container).

AleCandido:
@cschwan in https://github.com/NNPDF/pineappl/actions/runs/4344435880 we actually obtained all the wheels we wanted:
- we have all the full Python versions range ([3.7, 3.11]) for CPython for all the three platforms
- we have PyPy wheels for Linux

Moreover, the workflow is now running by default on `push:`, but without uploading the wheels (that should only happen on releases, but of course this I couldn't test).
Everything is documented in `update-wheels.sh`.

Should we agree on the current result, or do we have still some desiderata left behind?

AleCandido:
Curious fact: Windows has the smallest wheels, while MacOS the biggest ones

- Win: 676 KB
- Linux: 764 KB
- MacOS: 1.3 MB

The size of the wheel does not depend on the Python version or the interpreter (wheels for Linux-PyPy are the same size of Linux-CPython). This makes me believe that they are actually very similar, and only the platform makes an actual significant difference...

cschwan:
> Should we agree on the current result, or do we have still some desiderata left behind?

As far as it concerns me, that's perfect! Go ahead and merge it.

cschwan:
> Curious fact: Windows has the smallest wheels, while MacOS the biggest ones
> 
>     * Win: 676 KB
> 
>     * Linux: 764 KB
> 
>     * MacOS: 1.3 MB
> 
> 
> The size of the wheel does not depend on the Python version or the interpreter (wheels for Linux-PyPy are the same size of Linux-CPython). This makes me believe that they are actually very similar, and only the platform makes an actual significant difference...

These numbers make sense, because for MacOS the unversal wheels are basically two wheels in one, for x86_64 and aarch64.

PR #222, opened 2023-03-28
Sparse to dense

andreab1997:
We want a function that starting from a Sparsearray3 is able to provide a `ndarray` that can be then modified in python and converted to a Sparsearray again. @AleCandido

### DIFF
--- a/pineappl/src/import_only_subgrid.rs
+++ b/pineappl/src/import_only_subgrid.rs
@@ -3,7 +3,6 @@
 use super::grid::Ntuple;
 use super::sparse_array3::SparseArray3;
 use super::subgrid::{Mu2, Stats, Subgrid, SubgridEnum, SubgridIndexedIter};
-use ndarray::Array3;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::mem;
@@ -209,9 +208,6 @@ impl ImportOnlySubgridV2 {
         &mut self.array
     }
 
-    pub fn dense(& self) -> Array3<f64> {
-        self.array.to_ndarray()
-    }
 }
 
 impl Subgrid for ImportOnlySubgridV2 {
--- a/pineappl/src/import_only_subgrid.rs
+++ b/pineappl/src/import_only_subgrid.rs
@@ -3,6 +3,7 @@
 use super::grid::Ntuple;
 use super::sparse_array3::SparseArray3;
 use super::subgrid::{Mu2, Stats, Subgrid, SubgridEnum, SubgridIndexedIter};
+use ndarray::Array3;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::mem;
@@ -206,6 +207,10 @@ impl ImportOnlySubgridV2 {
     /// Return the array containing the numerical values of the grid.
     pub fn array_mut(&mut self) -> &mut SparseArray3<f64> {
         &mut self.array
+    }
+
+    pub fn dense(& self) -> Array3<f64> {
+        self.array.to_ndarray()
     }
 }
 
--- a/pineappl/src/import_only_subgrid.rs
+++ b/pineappl/src/import_only_subgrid.rs
@@ -207,7 +207,6 @@ impl ImportOnlySubgridV2 {
     pub fn array_mut(&mut self) -> &mut SparseArray3<f64> {
         &mut self.array
     }
-
 }
 
 impl Subgrid for ImportOnlySubgridV2 {
--- a/pineappl/src/sparse_array3.rs
+++ b/pineappl/src/sparse_array3.rs
@@ -428,16 +428,6 @@ impl<T: Clone + Default + PartialEq> SparseArray3<T> {
     }
 }
 
-impl<T: Copy + Default + PartialEq> SparseArray3<T>{
-    /// Cast to array.
-    pub fn to_ndarray(&self) -> Array3<T> {
-        let mut arr = Array::from_elem(self.dimensions, T::default());
-        for ((a,b,c), value) in self.indexed_iter(){
-            arr[[a,b,c]] = value; 
-        }
-        arr
-    }
-}
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/pineappl/src/sparse_array3.rs
+++ b/pineappl/src/sparse_array3.rs
@@ -310,7 +310,6 @@ impl<T: Clone + Default + PartialEq> SparseArray3<T> {
         }
     }
 
-
     /// Clear the contents of the array.
     pub fn clear(&mut self) {
         self.entries.clear();
--- a/pineappl/src/sparse_array3.rs
+++ b/pineappl/src/sparse_array3.rs
@@ -1,5 +1,6 @@
 //! Module containing the `SparseArray3` struct.
 
+use ndarray::{Array, Array3};
 use ndarray::{ArrayView3, Axis};
 use serde::{Deserialize, Serialize};
 use std::iter;
@@ -309,6 +310,7 @@ impl<T: Clone + Default + PartialEq> SparseArray3<T> {
         }
     }
 
+
     /// Clear the contents of the array.
     pub fn clear(&mut self) {
         self.entries.clear();
@@ -426,6 +428,16 @@ impl<T: Clone + Default + PartialEq> SparseArray3<T> {
     }
 }
 
+impl<T: Copy + Default + PartialEq> SparseArray3<T>{
+    /// Cast to array.
+    pub fn to_ndarray(&self) -> Array3<T> {
+        let mut arr = Array::from_elem(self.dimensions, T::default());
+        for ((a,b,c), value) in self.indexed_iter(){
+            arr[[a,b,c]] = value; 
+        }
+        arr
+    }
+}
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/pineappl/src/sparse_array3.rs
+++ b/pineappl/src/sparse_array3.rs
@@ -1,6 +1,5 @@
 //! Module containing the `SparseArray3` struct.
 
-use ndarray::{Array, Array3};
 use ndarray::{ArrayView3, Axis};
 use serde::{Deserialize, Serialize};
 use std::iter;
--- a/pineappl/src/sparse_array3.rs
+++ b/pineappl/src/sparse_array3.rs
@@ -376,9 +376,8 @@ impl<T: Clone + Default + PartialEq> SparseArray3<T> {
         } else if x < self.start + nx {
             let at = (x - self.start) * dim1;
             let offset = self.indices[at].1;
-            let dim2 = self.dimensions.1.max(self.dimensions.2);
             self.indices
-                .splice(at..at, iter::repeat((0, offset)).take(dim2));
+                .splice(at..at, iter::repeat((0, offset)).take(dim1));
         } else if x <= self.dimensions.0 {
             // nothing to do here
         } else {
@@ -1098,5 +1097,43 @@ mod tests {
         assert_eq!(other[[1, 8, 0]], 6.0);
         assert_eq!(other[[1, 9, 0]], 7.0);
         assert_eq!(other[[2, 0, 0]], 8.0);
+    }
+
+    // https://github.com/NNPDF/pineappl/issues/220
+    #[test]
+    fn regression_test_220() {
+        let mut array = SparseArray3::new(1, 2, 4);
+
+        array[[0, 0, 0]] = 1.0;
+
+        assert_eq!(array[[0, 0, 0]], 1.0);
+
+        assert_eq!(array.x_range(), 0..1);
+
+        let mut iter = array.indexed_iter();
+
+        assert_eq!(iter.next(), Some(((0, 0, 0), 1.0)));
+        assert_eq!(iter.next(), None);
+
+        array.increase_x_at(0);
+
+        array[[0, 0, 0]] = 2.0;
+
+        let mut iter = array.indexed_iter();
+
+        assert_eq!(iter.next(), Some(((0, 0, 0), 2.0)));
+        assert_eq!(iter.next(), Some(((1, 0, 0), 1.0)));
+        assert_eq!(iter.next(), None);
+
+        array.increase_x_at(1);
+
+        array[[1, 0, 0]] = 3.0;
+
+        let mut iter = array.indexed_iter();
+
+        assert_eq!(iter.next(), Some(((0, 0, 0), 2.0)));
+        assert_eq!(iter.next(), Some(((1, 0, 0), 3.0)));
+        assert_eq!(iter.next(), Some(((2, 0, 0), 1.0)));
+        assert_eq!(iter.next(), None);
     }
 }
--- a/pineappl/src/subgrid.rs
+++ b/pineappl/src/subgrid.rs
@@ -5,8 +5,8 @@ use super::grid::Ntuple;
 use super::import_only_subgrid::{ImportOnlySubgridV1, ImportOnlySubgridV2};
 use super::lagrange_subgrid::{LagrangeSparseSubgridV1, LagrangeSubgridV1, LagrangeSubgridV2};
 use super::ntuple_subgrid::NtupleSubgridV1;
-use ndarray::Array3;
 use enum_dispatch::enum_dispatch;
+use ndarray::Array3;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
@@ -77,8 +77,12 @@ pub trait Subgrid {
     /// return an empty slice.
     fn x2_grid(&self) -> Cow<[f64]>;
 
-    fn dimensions(& self) -> (usize, usize, usize) {
-        (self.mu2_grid().len(), self.x1_grid().len(), self.x2_grid().len())
+    fn dimensions(&self) -> (usize, usize, usize) {
+        (
+            self.mu2_grid().len(),
+            self.x1_grid().len(),
+            self.x2_grid().len(),
+        )
     }
 
     /// Convolute the subgrid with a luminosity function, which takes indices as arguments that
@@ -122,14 +126,13 @@ pub trait Subgrid {
     fn static_scale(&self) -> Option<Mu2>;
 
     /// Return the array
-    fn dense(& self) -> Array3<f64> {
+    fn dense(&self) -> Array3<f64> {
         let mut arr = Array3::from_elem(self.dimensions(), f64::default());
-        for ((a,b,c), value) in self.indexed_iter(){
-            arr[[a,b,c]] = value; 
+        for ((a, b, c), value) in self.indexed_iter() {
+            arr[[a, b, c]] = value;
         }
         arr
     }
-
 }
 
 /// Type to iterate over the non-zero contents of a subgrid. The tuple contains the indices of the
--- a/pineappl/src/subgrid.rs
+++ b/pineappl/src/subgrid.rs
@@ -77,15 +77,6 @@ pub trait Subgrid {
     /// return an empty slice.
     fn x2_grid(&self) -> Cow<[f64]>;
 
-    /// Return the dimension of the dense array of the subgrid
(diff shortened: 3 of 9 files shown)


AleCandido:
In particular, we'd like to have this function exposed by `ImportOnkySubgridV2`, since they are the result of the optimization, in order to be able to use it to export, modify, and reimport subgrids.

cschwan (changes requested):
I believe we can make this more general and implement it as

```rust
impl From<&SubgridEnum> for Array3<f64> {
    fn from(&SubgridEnum) -> Self {
        // content of `to_ndarray`
    }
}
```

Also, please run `cargo fmt` to format the source code properly. If that doesn't work install it using `rustup component add rustfmt`.

AleCandido:
> I believe we can make this more general and implement it as

Even fine, I like using stdlib traits, they make everything more idiomatic. But for most `SubgridEnum` kinds we should just fail, isn't it?

cschwan:
You do not need distinguish between the various types because `SubgridEnum` implements the `Subgrid` trait, which in turn has `Subgrid::indexed_iter()` returning the same iterator that `ndarray::Array3` and `SparseArray3` return. So if you take my comment in the review literally it should work (and that's what meant).

AleCandido:
> You do not need distinguish between the various types because `SubgridEnum` implements the `Subgrid` trait, which in turn has `Subgrid::indexed_iter()` returning the same iterator that `ndarray::Array3` and `SparseArray3` return. So if you take my comment in the review literally it should work (and that's what meant).

Sorry, I interpreted to much: I was biased thinking that only `*Sparse*` subgrids were based on `SparseArray3`, and `IndexedIter` was only implemented by it (the second statement being wrong).

Then it is perfectly fine, the implementation is fully general, and we can move it to `subgrid.rs`.

@andreab1997 this will also save some troubles, because you won't need to cast a `SubgridEnum` to `ImportOnlySubgridV2` to get the dense array.
We'll still need the opposite cast: we can't create a generic `SubgridEnum`, we have to create a specific one and fill with the array content, but then we'll need to cast it as `SubgridEnum` to fill the grid.
However, it is already one cast less, and the only one that we need is already implemented (for V1, but we just need to do the same for V2).
https://github.com/NNPDF/pineappl/blob/0564551a2bab88b07880665abec1cfb6f3f7efcc/pineappl_py/src/import_only_subgrid.rs#L60-L63

pineappl/src/import_only_subgrid.rs
```diff
@@ -207,6 +207,7 @@ impl ImportOnlySubgridV2 {
     pub fn array_mut(&mut self) -> &mut SparseArray3<f64> {
         &mut self.array
     }
+
```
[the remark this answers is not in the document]
AleCandido: In principle `cargo fmt` should also solve this, and related ones :)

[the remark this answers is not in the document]
andreab1997: [c741283](https://github.com/NNPDF/pineappl/pull/222/commits/c74128328a37176a6dda11f26bfcd093acaf72f9)

pineappl/src/sparse_array3.rs
```diff
@@ -309,6 +310,7 @@ impl<T: Clone + Default + PartialEq> SparseArray3<T> {
         }
     }
 
+
```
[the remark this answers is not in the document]
andreab1997: [c741283](https://github.com/NNPDF/pineappl/pull/222/commits/c74128328a37176a6dda11f26bfcd093acaf72f9)

pineappl/src/sparse_array3.rs
```diff
@@ -1,5 +1,6 @@
 //! Module containing the `SparseArray3` struct.
 
+use ndarray::{Array, Array3};
```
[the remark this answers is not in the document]
andreab1997: [c9372a3](https://github.com/NNPDF/pineappl/pull/222/commits/c9372a37fd1e09559540f77938e70d5dee973405)

pineappl_py/src/subgrid.rs
```diff
@@ -170,5 +173,12 @@ impl PySubgridEnum {
         self.clone()
     }
 
+    pub fn mu2_ren_grid<'py>(&self, py: Python<'py>) -> &'py PyArray1<f64> {
+        PyArray1::from_vec(py, self.subgrid_enum.mu2_grid().iter().map(|&Mu2 { ren, .. }| ren).collect())
+    }
+
+    pub fn mu2_fact_grid<'py>(&self, py: Python<'py>) -> &'py PyArray1<f64> {
+        PyArray1::from_vec(py, self.subgrid_enum.mu2_grid().iter().map(|&Mu2 { ren , fac }| fac).collect())
+    }
```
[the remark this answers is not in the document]
andreab1997: I agree but @cschwan and I tried and failed :)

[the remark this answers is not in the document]
AleCandido: Failed?

In case, I'll do myself asap

AleCandido:
@andreab1997 I should have added the relevant API for `Mu2`.

Can you test it?
Since `Mu2` is a non-trivial object (not a Python one), I preferred returning a normal list, instead of messing more with NumPy arrays (other non-Python objects, in a different way).
Of course, we can add wrappers in the pure Python layer, to simplify common operations, like:
```py
class SubgridEnum:
    def facgrid(self):
        return np.array([mu2.fac for mu2 in self.mu2_grid])
```
This would be 100% equivalent to what you were doing, but it clearly separates the thin PineAPPL wrapper and the Python-idiomatic code, such that, for performance’s sake, you can always access the raw methods.

E.g. you can call:
```py
ImportOnlySubgridV2.new_import_only_subgrid(array, old.mu2_grid(), old.x1_grid(), old.x2_grid())
```
For this reason I'd propose to also change the parameter type of `mu2grid` in `new_import_only_subgrid` from `Vec<(f64, f64)>` to `Vec<PyMu2>`, i.e. exactly what you get from a `SubgridEnum`.

AleCandido:
@andreab1997 to prove everything is working, can you also add suitable tests in `pineappl_py/tests/`?

Moreover, for further refinement, we should add Python wrappers for explicitly used new objects (i.e. `ImportOnlySubgridV2` should be defined in `pineappl_py/pineappl/import_only_subgrid.py`, but I don't see a pressing need for a `Mu2` python class, since that object is mostly internal).

pineappl/src/subgrid.rs
```diff
@@ -76,6 +77,14 @@ pub trait Subgrid {
     /// return an empty slice.
     fn x2_grid(&self) -> Cow<[f64]>;
 
+    fn dimensions(&self) -> (usize, usize, usize) {
```
AleCandido, line 80: Missing docstring.

andreab1997:
> @andreab1997 to prove everything is working, can you also add suitable tests in `pineappl_py/tests/`?
> 
> Moreover, for further refinement, we should add Python wrappers for explicitly used new objects (i.e. `ImportOnlySubgridV2` should be defined in `pineappl_py/pineappl/import_only_subgrid.py`, but I don't see a pressing need for a `Mu2` python class, since that object is mostly internal).
> 
> And we are missing some docstrings for the new elements. I added the warnings in `pineappl_py/src/lib.rs`, so even `maturin develop` should tell you where.

Yes I will do, thanks!

pineappl/src/subgrid.rs
```diff
@@ -115,6 +124,15 @@ pub trait Subgrid {
 
     /// Return the static (single) scale, if this subgrid has one.
     fn static_scale(&self) -> Option<Mu2>;
+
+    /// Return the array
```
AleCandido, line 128: I know the name is good enough on its own, but like this, this docstring is not very informative...
```suggestion
    /// Fill a dense array with subgrid content
```

pineappl_py/pineappl/import_only_subgrid.py
```diff
@@ -24,3 +25,24 @@ def __init__(self, array, q2_grid, x1_grid, x2_grid):
         self._raw = PyImportOnlySubgridV1(
             np.array(array), np.array(q2_grid), np.array(x1_grid), np.array(x2_grid)
         )
+
+class ImportOnlySubgridV2(PyWrapper):
```
AleCandido, line 29: Ah sorry, you already did it, ignore my other comment.

pineappl_py/src/subgrid.rs
```diff
@@ -162,7 +195,27 @@ impl PySubgridEnum {
         self.subgrid_enum.scale(factor);
     }
 
+    pub fn dense<'py>(&self, py: Python<'py>) -> &'py PyArray3<f64> {
```
AleCandido, line 198: Missing docs, and similarly the others.

pineappl_py/pineappl/import_only_subgrid.py
```diff
@@ -24,3 +25,24 @@ def __init__(self, array, q2_grid, x1_grid, x2_grid):
...
+    """
+    Python wrapper object to :class:`~pineappl.pineappl.PyImportOnlySubgridV2`.
+
+    Parameters
+    ----------
+        array : numpy.ndarray(float, dim=3)
+            3-dimensional subgrid content
+        mu2_grid : sequence(float)
```
AleCandido, line 37: `mu2_grid` can't be a sequence of floats, unless you do some magic later on, e.g.
```py
mu2s = [(mu2, mu2) for mu2 in mu2_grid]
```

But you'd need to account for both the options, so, as a first shot, better to accept a sequence of tuples. In fact, a sequence of `PyMu2` objects.

If you want a simpler constructor, allowing for a sequence of tuples, add a `@classmethod`, such that is always possible to construct from the raw object, in case you already have it.

andreab1997:
> @andreab1997 to prove everything is working, can you also add suitable tests in `pineappl_py/tests/`?
> 
> Moreover, for further refinement, we should add Python wrappers for explicitly used new objects (i.e. `ImportOnlySubgridV2` should be defined in `pineappl_py/pineappl/import_only_subgrid.py`, but I don't see a pressing need for a `Mu2` python class, since that object is mostly internal).
> 
> And we are missing some docstrings for the new elements. I added the warnings in `pineappl_py/src/lib.rs`, so even `maturin develop` should tell you where.

I tested a bit and everything seems to work. I will soon add some tests.

pineappl_py/src/subgrid.rs
```diff
@@ -195,26 +195,27 @@ impl PySubgridEnum {
...
+    /// Return the dense array of the subgrid.
     pub fn dense<'py>(&self, py: Python<'py>) -> &'py PyArray3<f64> {
         self.subgrid_enum.dense().into_pyarray(py)
     }
 
     pub fn into(&self) -> Self {
         self.clone()
     }
-
+    /// Return the array of mu2 objects of a subgrid
```
AleCandido, line 206: When you start with "Return" consider replacing with something more informative (all functions return something, or at least they should).

In this case, we could consider just dropping it, and turn the method into a more Pythonic `@property` (PyO3 expose supports this as `#[getter]`).
https://pyo3.rs/v0.18.2/class.html#object-properties-using-getter-and-setter

pineappl_py/tests/test_sugrid.py
```diff
@@ -28,3 +30,44 @@ def convolute_grid():
...
+    def test_dense(self):
+        grid = self.fake_grid()
+        test_subgrid, infos = self.fake_importonlysubgrid()
+        _, _, _, array = (obj for obj in infos)
+        grid.set_subgrid(0,0,0, test_subgrid)
+        extr_subgrid = grid.subgrid(0,0,0)
+        test_array = extr_subgrid.dense()
+        print(test_array)
+        print(array)
+        np.testing.assert_allclose(test_array, array)
```
cschwan: <https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b>

pineappl_py/src/subgrid.rs
```diff
@@ -142,6 +143,53 @@ impl PySubgridParams {
     }
 }
 
+/// PyO3 wrapper to :rustdoc:`pineappl::subgrid::Mu2 <subgrid/struct.Mu2.html>`
+#[pyclass]
+#[repr(transparent)]
+pub struct PyMu2 {
+    pub(crate) mu2: Mu2,
```
cschwan, line 152: Is `pub(crate)` really needed?

pineappl_py/src/subgrid.rs
```diff
@@ -142,6 +143,53 @@ impl PySubgridParams {
...
+}
+impl PyMu2 {
+    pub(crate) fn new(mu2: Mu2) -> Self {
```
cschwan, line 155: Drop this block ...

pineappl_py/src/subgrid.rs
```diff
@@ -142,6 +143,53 @@ impl PySubgridParams {
...
+        Self { mu2 }
+    }
+}
+
+#[pymethods]
+impl PyMu2 {
+    #[new]
+    pub fn default(ren: f64, fact: f64) -> Self {
```
cschwan, line 163: ... and rename this function `new` instead, constructing the return value immediately:

```rust
Self { mu2: Mu2 { ren, fac } }
```

pineappl/src/subgrid.rs
```diff
@@ -32,7 +33,7 @@ pub enum SubgridEnum {
 }
 
 /// Structure denoting renormalization and factorization scale values.
-#[derive(Debug, Deserialize, Clone, PartialEq, PartialOrd, Serialize)]
+#[derive(Debug, Deserialize, Clone, Copy, PartialEq, PartialOrd, Serialize)]
```
cschwan, line 36: Why/where do you need `Copy`?

pineappl_py/src/import_only_subgrid.rs
```diff
@@ -2,10 +2,75 @@ use super::subgrid::PySubgridEnum;
...
+        Self {
+            import_only_subgrid,
+        }
+    }
+}
+
+#[pymethods]
+impl PyImportOnlySubgridV2 {
+    #[new]
+    pub fn new_import_only_subgrid(
```
cschwan, line 31: Call this constructor simply `new`.

pineappl_py/src/import_only_subgrid.rs
```diff
@@ -2,10 +2,75 @@ use super::subgrid::PySubgridEnum;
...
+///
+/// **Usage**: `pineko`
+#[pyclass]
+#[derive(Clone)]
+#[repr(transparent)]
+pub struct PyImportOnlySubgridV2 {
+    pub(crate) import_only_subgrid: ImportOnlySubgridV2,
+}
+
+impl PyImportOnlySubgridV2 {
```
cschwan, line 20: Remove this block.

pineappl_py/src/import_only_subgrid.rs
```diff
@@ -2,10 +2,75 @@ use super::subgrid::PySubgridEnum;
...
+        ))
+    }
+
+    /// Wrapper to match :meth:`pineappl.pineappl.PyGrid.set_subgrid()`
+    ///
+    /// Returns
+    /// -------
+    ///     PySubgridEnum :
+    ///         casted object
+    pub fn into(&self) -> PySubgridEnum {
```
cschwan, line 67: This function seems to do nothing but clone itself, try to remove it.

cschwan:
I addressed <https://github.com/NNPDF/pineappl/pull/222#pullrequestreview-1365213961> in commit 0504a720f34e10cbc58ae17ba5cbcbd0ada61717.

pineappl_py/tests/test_sugrid.py
cschwan: I added a newline in commit 0504a720f34e10cbc58ae17ba5cbcbd0ada61717.

cschwan (changes requested):

pineappl_py/src/import_only_subgrid.rs
AleCandido: Not really: this function is a recast.

pineappl_py/src/import_only_subgrid.rs
AleCandido: Fine: since the attribute is `pub(crate)` anyhow, you do not need another `pub(crate)` constructor.

pineappl_py/src/import_only_subgrid.rs
cschwan: Right!

pineappl_py/src/subgrid.rs
AleCandido: No, since you have `{set_}{ren,fac}` accessors down below, you do not need any other way to access.

pineappl_py/src/subgrid.rs
andreab1997: [e557df8](https://github.com/NNPDF/pineappl/pull/222/commits/e557df89a81f2a1b225deeae42df854955d3bc3c)

pineappl_py/src/subgrid.rs
andreab1997: [e557df8](https://github.com/NNPDF/pineappl/pull/222/commits/e557df89a81f2a1b225deeae42df854955d3bc3c)

pineappl_py/src/subgrid.rs
andreab1997: [ce61075](https://github.com/NNPDF/pineappl/pull/222/commits/ce610751d38c27a588a0c8592579186d89497bc0)

pineappl/src/subgrid.rs
andreab1997: If I remove it I get the following error:
```
error[E0507]: cannot move out of a shared referenceineappl_py                                                                                                         
   --> pineappl_py/src/subgrid.rs:219:19
    |
219 |             .map(|&mu2| PyMu2 { mu2 })
    |                   ^---
    |                   ||
    |                   |data moved here
    |                   |move occurs because `mu2` has type `Mu2`, which does not implement the `Copy` trait
```

cschwan: In that case use say `.cloned()` before the `.map` call. You probably have to change `|&mu2|` to `|mu2|` (or `move |mu2|` ?).

andreab1997: Ok thanks! [9fb4e10](https://github.com/NNPDF/pineappl/pull/222/commits/9fb4e105e7f15d11b9016ed81555bd3218a2ea79)

pineappl_py/src/import_only_subgrid.rs
andreab1997: If I do this it complains that there are two methods called new. I believed I have done some mess with the declarations above

cschwan: I should've been clearer:

1. https://github.com/NNPDF/pineappl/pull/222/files#r1179250872
2. then rename the constructor to `new`. At the very least you should get another error message (but hopefully none).

cschwan: Try replacing this line with

```rust
Self {
    import_only_subgrid: ImportOnlySubgridV2::new(
        // ..
    )
}
```

The `Self { ... }` syntax is how you construct an object if you have access to its members (if you're in the same module or if all members are visible to you).

pineappl_py/src/import_only_subgrid.rs
andreab1997: [605fd90](https://github.com/NNPDF/pineappl/pull/222/commits/605fd90d088b4c6a384dac6b0c162460328393e7)

pineappl_py/src/import_only_subgrid.rs
andreab1997: Thanks! Addressed in [605fd90](https://github.com/NNPDF/pineappl/pull/222/commits/605fd90d088b4c6a384dac6b0c162460328393e7)

andreab1997:
@cschwan I am asking again your review just to be sure that everything is ready to be merged for you

cschwan (approved):
Looks good to me!

pineappl_py/src/subgrid.rs
cschwan: I think this is OK for the time being.

Issue #226, opened 2023-04-20
`Grid::optimize` causes `Grid::evolve` to produce wrong `FkTable`s


andreab1997:
Computing FKs with `pineappl==0.6.a5` and comparing their predictions with the predictions that can be obtained from the grids (i.e. using `pineko compare`) results in some of the FK predictions to be off by up to a 100% factor. Considering all the hadronic FKs in NNPDF40 datasets, most of them are fine but there are a (not negligible) few which are off.

This is the comparison I mentioned  
[comparison_424.txt](https://github.com/NNPDF/pineappl/files/11287680/comparison_424.txt)

Just skipping the grid optimization in `pineko` (i.e. `grid.optimize()`) the problem disappears.

Here is the same comparison but with the new FKs
[comparison_424.txt](https://github.com/NNPDF/pineappl/files/11287685/comparison_424.txt)

To simplify the debug process, here is the list of the mostly affected dataset:

- ATLAS_WP_JET_8TEV_PT
- ATLAS_WM_JET_8TEV_PT
- ATLASZPT8TEVMDIST
- ATLASZPT8TEVYDIST
- ATLAS_TTB_DIFF_8TEV_LJ_TRAPNORM
- ATLAS_TTB_DIFF_8TEV_LJ_TTRAPNORM
- ATLAS_TOPDIFF_DILEPT_8TEV_TTRAPNORM
- CMSZDIFF12
- CMS_TTBAR_2D_DIFF_MTT_TRAP_NORM
- CMS_TTB_DIFF_13TEV_2016_2L_TRAP
- CMS_TTB_DIFF_13TEV_2016_LJ_TRAP

which of course can be inferred by the comparisons I posted. 

@cschwan @AleCandido @felixhekhorn 

As far as I know at the moment only me, @niclaurenti and @giacomomagni are computing fktables. I tag you here so that you are aware and you can check your FKs as well.

felixhekhorn:
Note that for
- `ATLAS_WP_7TEV` the two runs are digit by digit the same and good and this is re-computed, not converted
- `ATLAS_WP_JET_8TEV_PT-atlas-atlas-wjets-arxiv-1711.03296-xsec002` the thing went wrong as @andreab1997 said and it is converted
- `ATLASWZRAP36PB-ATLAS-arXiv:1109.5141-Wminus_eta3` both versions are fine, but *not* the same digit by digit and it is converted

[...]

PR #227, opened 2023-04-21
Support EKO's new file format

cschwan:


cschwan:
@AleCandido @felixhekhorn could one of you have a look at commit a735ccf8691c2a244f21724db9fe05454714d265 and tell me if I'm parsing `metadata.yaml` correctly and if there are any obvious mistakes? When I use the code it doesn't work for the specific example I'm considering.

pineappl_cli/src/evolve.rs
```diff
@@ -37,28 +32,124 @@ fn evolve_grid(
         targetpids: Vec<i32>,
     }
 
-    let mut archive = Archive::new(File::open(eko)?);
+    #[derive(Default, Deserialize)]
+    struct Rotations {
+        #[serde(rename = "_targetgrid")]
```
AleCandido, line 37: `_targetgrid` is saved without a prepended underscore.

Essentially, the conversion you're doing here is already done while saving.

AleCandido: Unfortunately, this has recently changed...

AleCandido:
> In particular, I'm unsure whether I'm handling the PIDs correctly. There are `_inputpids`, `_targetpids`, `pids` and I don't know what it means if they're set to `null` and how to interpret `_inputpids`.

If they are `null`, `*pids` are equal to `pids`, and `*grid` are equal to `xgrid`.

pineappl_cli/src/evolve.rs
cschwan: In the `metadata.yaml` file that I'm looking at the format is:

```yaml
rotations:
  _targetgrid:
```

AleCandido:
These are all internal details of the EKO format, and we didn't consider that PineAPPL was actually parsing EKOs independently.

Because of this, I would move soon the Rust parser into EKO, such that we can keep updating the two consistently (even though we'd like to not update the format too frequently).

cschwan:
> These are all internal details of the EKO format, and we didn't consider that PineAPPL was actually parsing EKOs independently.
> 
> Because of this, I would move soon the Rust parser into EKO, such that we can keep updating the two consistently (even though we'd like to not update the format too frequently).

I agree, this is better. For the time being I'm mainly interested in this to debug #226, of course.

pineappl_cli/src/evolve.rs
AleCandido: Ok, then you got an old EKO...

https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/eko/io/runcards.py#L206-L209

AleCandido:
> I agree, this is better. For the time being I'm mainly interested in this to debug #226, of course.

I know, and I agree. Debug is more urgent than improvements.

When I said soon™ I didn't mean *now* ;D

pineappl_cli/src/evolve.rs
cschwan: Oh I see, but no worries, as you can see `serde` allows to very easily support different formats.

cschwan: Commit 5378a41d020c6db9f16030bc0b4fe60d41b1de31 should support both (as the line you pointed out in EKO). I continue to be impressed by serde!

cschwan:
I think the problem is that the EKO I have uses the evolution basis, and `inputpids` specifies the evolution basis as a linear combination of the flavor basis. I that right @AleCandido?

AleCandido:
> I think the problem is that the EKO I have uses the evolution basis, and `inputpids` specifies the evolution basis as a linear combination of the flavor basis. Is that right @AleCandido?

Yes, we used for some time a "fully generic approach", to support a whatever flavor basis. To do this we had to use the rotation matrices.

However, this is sometimes a problem, because we're mostly used to work with PIDs lists, so, when propagated, it gives a lot of troubles.
If I remember correctly, we decided with @felixhekhorn and @andreab1997 to stop doing this, and simplify to the only basis we're using externally (i.e. the same of APFEL: canonical, evolution, and unified evolution).

But I'm not sure to which level this decision has been propagated (throughout EKO, to Pineko, and so on).

cschwan:
What I'm seeing listed in `inputpids:` in my `metadata.yaml` is a list of factors that exactly specify the evolution basis. Changing `pids0` and `lumi_id_types` accordingly in the contruction of `OperatorInfo` in the `Metadata::V1` matching arm the code above then works, so that means I can debug #226 now.

cschwan:
> If I remember correctly, we decided with @felixhekhorn and @andreab1997 to stop doing this, and simplify to the only basis we're using externally (i.e. the same of APFEL: canonical, evolution, and unified evolution).

Would that mean that `inputpids:` would list the PIDs of the evolution basis, similar to `pids`?

AleCandido:
> Would that mean that `inputpids:` would list the PIDs of the evolution basis, similar to `pids`?

Exactly :)

cschwan:
OK, great. One (next-to-?) final question: are all of `inputgrid`, `inputpids`, `targetgrid` and `targetpids` allowed to be `null` and if so, these values are taken from `xgrid` and `pids`, respectively?

andreab1997:
> OK, great. One (next-to-?) final question: are all of `inputgrid`, `inputpids`, `targetgrid` and `targetpids` allowed to be `null` and if so, these values are taken from `xgrid` and `pids`, respectively?

Basically when computing the eko, all these parameters are not even used even if they are in the operator card. So the eko is always computed with an interpolation grid given by `xgrid` and with `pids` given by `pids`. Then you can rotate the eko in these spaces and, if you do, `targetgrid`, `inputgrid` and so on will be updated in the correct way.

cschwan:
But what if all of these values are the same, so if there's no rotation, for instance if you generate an EKO for flavor-to-flavor basis, where at each scale the x-grid is the same? I'm asking whether I should allow `null` in the yaml parser in `pineappl evolve`.

PR #229, opened 2023-05-10
Upload libraries for Linux

cschwan:
- [ ] where can the library be used? Make the requirements explicit
- [ ] do we have to patch the `prefix` field in the `.pc` file? If yes, we need an installer that does it.

### DIFF
--- a/.github/workflows/capi.yaml
+++ b/.github/workflows/capi.yaml
@@ -1,6 +1,6 @@
 name: CAPI
 
-on: push
+#on: push
 
 jobs:
   capi:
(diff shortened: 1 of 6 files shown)


AleCandido:
In principle, you can add the compiled library also as an attachment to releases, from which it would be simpler to collect even for external users.

cschwan:
@AleCandido that's the idea, eventually.

I'm currently still experimenting a bit with the binaries to see whether they're really usable everywhere. I suspect they're not, depending on the (g)libc versions available during compilation/execution and some other paths and OS/distribution-specific things that I havent't figured out yet. Another problem is that the `.pc` file has its installation path hardcoded into it, meaning the user must change the `prefix` accordingly. For that we probably want some small installation script asking for the installation directory. I'm thinking of something similar to what [rustup](https://rustup.rs/) is using.

PR #239, opened 2023-08-16
Support creation of wheels for Python 3.6

cschwan:
Because of popular requests we may consider adding support for Python 3.6.


AleCandido:
The changes look alright, in principle, but they should be tested. Have you managed to do it by chance?

Otherwise, we could try to make a *dry-run* mode of the workflow, where no upload is happening, and manually trigger it (there is the `workflow-dispatch` event for this, in case you were not aware).

AleCandido:
> That's what I did in commit [9d29741](https://github.com/NNPDF/pineappl/commit/9d297412ad40c47188523deb47efe68ae1fe2cfc), and then the actions have to be triggered manually, as you said.

Ok, great. I was not aware, you're just ahead of me ;)

AleCandido:
Unfortunately, the repository contains `/` characters. I'd suggest you replacing `/` with `@` in your `sed` command.

.github/workflows/release.yml
```diff
@@ -146,6 +146,48 @@ jobs:
...
+          # overwrite it explicitly
+          sed -i \
+              -e "s/categories.workspace = true/$(grep 'categories' Cargo.toml)/" \
+              -e "s/edition.workspace = true/$(grep 'edition' Cargo.toml)/" \
+              -e "s/keywords.workspace = true/$(grep 'keywords' Cargo.toml)/" \
+              -e "s/license.workspace = true/$(grep 'license' Cargo.toml)/" \
+              -e "s/repository.workspace = true/$(grep 'repository' Cargo.toml)/" \
+              -e "s/rust-version.workspace = true/$(grep 'rust-version' Cargo.toml)/" \
+              -e "s/version.workspace = true/$(grep 'version' Cargo.toml)/" \
+              pineappl_py/Cargo.toml
```
AleCandido, line 173:
```suggestion
          sed -i \
              -e "s@categories.workspace = true@$(grep 'categories' Cargo.toml)@" \
              -e "s@edition.workspace = true@$(grep 'edition' Cargo.toml)@" \
              -e "s@keywords.workspace = true@$(grep 'keywords' Cargo.toml)@" \
              -e "s@license.workspace = true@$(grep 'license' Cargo.toml)@" \
              -e "s@repository.workspace = true@$(grep 'repository' Cargo.toml)@" \
              -e "s@rust-version.workspace = true@$(grep 'rust-version' Cargo.toml)@" \
              -e "s@version.workspace = true@$(grep 'version' Cargo.toml)@" \
              pineappl_py@Cargo.toml
```

cschwan:
> Unfortunately, the repository contains `/` characters. I'd suggest you replacing `/` with `@` in your `sed` command.

Good catch :+1:!

AleCandido (approved):
It seems working, so I guess you could merge when you prefer.

I haven't tested the wheels, but with the latest layout existing wheels are unaltered, so you should try to download and see if the new (old) ones are good enough for the user purposes.

cschwan:
It seems to work since commit 2d632d3b440c58e4909c8a1518ba07e7e986ffc9, however the output is different, probably due to a different `maturin` version being used. @AleCandido please have a look at the wheels in <https://github.com/NNPDF/pineappl/actions/runs/5880630299>. Instead of a wheel for Python 3.6 we have a tar.gz file there. Do you know what it is?

.github/workflows/release.yml
cschwan: Fixed in commit 2d632d3b440c58e4909c8a1518ba07e7e986ffc9.

AleCandido:
@cschwan the `.tar.gz` is a source distribution, it's not compiled. There is actually no wheel for py3.6.

AleCandido:
The CPython 3.6 wheel has never been compiled, not only not uploaded. You can check by yourself searching for `Built wheel for` in the logs search box.
image

Actually, thinking further about the compilation process, it is happening inside a Manylinux container in any case, so the host system is irrelevant: the Python versions discovered are those available in the container in any case.
It seems like the version of Maturin used for EOL (that is the same one we used for long time) is using a container with CPython 3.7-10 and PyPy 3.7. In order to support CPython 3.6 we need to use a Manylinux container that has it installed, or install ourselves inside the container (but it is managed by the `maturin-action`...).

AleCandido:
However, you could customize `maturine-action` passing a Manylinux tag:

https://github.com/PyO3/maturin-action#inputs

Now the game is finding one with CPython 3.6 installed...

cschwan:
This doesn't seem to be a problem, everything seems to work now: <https://github.com/NNPDF/pineappl/actions/runs/5887659565>.

AleCandido:
> This doesn't seem to be a problem, everything seems to work now: [NNPDF/pineappl/actions/runs/5887659565](https://github.com/NNPDF/pineappl/actions/runs/5887659565).

You're right, even though I'm not sure how it is compiling for py3.6 if not present...
But I'm not an expert of writing manually Python C extensions (i.e. manually compiling with Python headers and the Python loading mechanism), so I'm missing something for sure.

PR #248, opened 2023-10-24
Add double ratio plot

andreab1997:
Adding a plot with the ratio of the first and third PDFs over the first and of the second and fourth PDFs over the second.


cschwan:
Some of the tests will probably fail, but I'll fix them myself. The patches look good to me!

PR #256, opened 2024-02-02
Complete the Fortran interface and extend the Python interface

janw20:
Hi, I was working on using PineAPPL with an existing Fortran code, and during that, I completed the Fortran interface to the PineAPPL C API, wrote a file to test this and added an example how to use the interface with LHAPDF, and also I implemented the convolute_with_two function in the Python interface.


cschwan:
Hi @janw20, thank you for your pull request! I had a look at your changes and they look pretty much as they should. I will comment on a few minor details, but after fixing them we can merge the PR quite soon.

examples/fortran/lhapdf_example.f90
```diff
@@ -0,0 +1,105 @@
...
+        real(c_double)                    :: alphas_test2
+
+        integer, pointer :: flags(:)
+
+        call c_f_pointer(state, flags, [2])
+
+        call lhapdf_alphasq2(0, 0, q2, alphas_test2)
+    end function
+
+end program lhapdf_example
```
cschwan, line 105: Please add a [final newline](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) here, which should make Github's 'stop sign' disappear.

examples/fortran/test.f90
```diff
@@ -0,0 +1,201 @@
...
+        implicit none
+        
+        real(c_double), value, intent(in) :: q2
+        type(c_ptr), value, intent(in)    :: state
+        real(c_double)                    :: alphas_test
+
+        alphas_test = q2
+    end function
+
+end program test_pineappl
```
cschwan, line 201: Also here a new line character is missing.

cschwan:
We probably also want to improve the following points (not necessarily in this PR):

- include the new program in our CI and include `pineappl.f90` into code coverage
- install the `*.mod` along with the rest of the CAPI; for why this isn't trivial see the discussion and pointers here: <https://fortran-lang.discourse.group/t/installing-fortran-module-files/1135/5>

janw20:
Hi @cschwan, thanks for your comments, I added the terminating newlines. I started learning Rust but I'm not very proficient yet, so I don't know if I can contribute to how the Fortran module would be compiled with cargo-c. But I think it would be nice if it would be installed together with the C API.

alecandido:
@cschwan is certainly more expert than myself (he did it) but, ttbomu,the only two things you should be concerned about for C installation are:

https://github.com/NNPDF/pineappl/blob/11cc43d2f54e0046688210a004dedcab26fc5958/pineappl_capi/Cargo.toml#L25-L26
and
https://github.com/NNPDF/pineappl/blob/master/pineappl_capi/cbindgen.toml

They are both declarative and pretty much straightforward.

For Fortran I guess there are additional problems, not related to Rust or `cargo-c`, but to Fortran itself. Most likely, the discussion pointed out by @cschwan is a great starting point (and my guess is that the main issue is the portability of the possible alternatives...)

cschwan:
No worries, @janw20, my comment above was more about writing down the ideas so they don't get lost. I created dedicated Issues for them now: #257 and #258.

cschwan:
This looks good and I'm merging it now. Thanks again, @janw20!

PR #273, opened 2024-04-15
Expose the luminosity and bin mask options for FKs to the python interface

scarlehoff:
It _seems_ to work. I've just copied it from the grid.

### DIFF
--- a/pineappl_py/pineappl/fk_table.py
+++ b/pineappl_py/pineappl/fk_table.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from .pineappl import PyFkTable, PyFkAssumptions
 from .utils import PyWrapper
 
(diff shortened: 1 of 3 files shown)


cschwan:
This looks good to me. It would be perfect if you could also add a small test.

pineappl_py/pineappl/fk_table.py
```diff
@@ -50,6 +52,42 @@ def optimize(self, assumptions = "Nf6Ind"):
...
+                cross sections for all bins, for each scale-variation tuple (first all bins, then
+                the scale variation)
+        """
+        return self.raw.convolute_with_one(
+            pdg_id,
+            xfx,
+            np.array(bin_indices),
+            np.array(lumi_mask),
+        )
+
```
alecandido, line 90: This is unnecessary. The `PyWrapper.__getattr__` is already exposing it, from the underlying object. The `numpy` crate is doing the NumPy-to-ndarray conversion.

scarlehoff: Which part is unnecessary? The `np.array`?

alecandido: Then read the docs :P

https://pyo3.rs/v0.21.2/function/signature#using-pyo3signature--

alecandido: I proposed something in c42be03, but I haven't tested yet (especially the value assigned to the defaults) because I still have to locally rebase on #245.

alecandido:
I'm working on fixing the signature.

Of course, you could drop the last commit and merge, but I believe to be valuable to avoid some overhead (whenever possible).

scarlehoff:
As you wish, since you are at it please add also the test that @cschwan asked for.

scarlehoff:
Thanks @alecandido 

As a note for the future, we should add maybe a test that does the entire `grid` -> `eko-convolution` -> `fktable` and checks that the `fktable` can be read. Just to get a heads up in case any part of the interface changes.

PR #275, opened 2024-04-18
Add new type `PackedArray`

janw20:


alecandido:
If it's packed, why not using directly `ndarray`?

https://docs.rs/ndarray/latest/ndarray/

It is standard, and widely supported. And `pineappl` already depends on it:
https://github.com/NNPDF/pineappl/blob/bc1de5b0669325ba8dcd8a4844373fe4acc302b6/pineappl/Cargo.toml#L29

cschwan:
`PackedArray` is possibly a bad name and that would be my fault. If you can think of a better name let us know!

The idea is to have a similar interface as the array class from `ndarray`, but not store zeros to keep down the file size, so `PackedArray<T, D>` will be the successor of`SparseArray3<T>`. The new type will compress a bit better and more importantly it supports arbitrary dimensions.

cschwan:
No worries! We're happy you're interested and that you're looking at the code.

alecandido:
Btw: what about just using `SparseArray`? (w/o the 3)

cschwan:
The adjective 'sparse' really never made sense, because when you'd have to decide whether the arrays were sparse or dense, they're clearly dense.

Another way of explaining it is that the algorithms for sparse matrices would clearly fail for use cases and in fact produce larger data structures.

alecandido:
From your comment, I'd say that it is an optimized strategy for the sparse matrix representation (as there is not a single best strategy). I.e., your matrix has a relevant amount of *default* elements, and you could compress your representation by avoiding writing down all elements.

However, names are relevant but not crucial. So, if *packed* sounds more appropriate, just keep that :)

cschwan:
> From your comment, I'd say that it is an optimized strategy for the sparse matrix representation (as there is not a single best strategy). I.e., your matrix has a relevant amount of _default_ elements, and you could compress your representation by avoiding writing down all elements.

That's basically it, we first map the `D`-dimensional indices to one-dime sional and then we store the non-zero (non-defaulted) elements a linearized vector. For zeros we just save where they are and how many of them are consecutive. Finally there's an exception where it makes sense to explicitly store zeros, however in practice it shouldn't make a big difference.

janw20:
I wrote 4 benchmarks with the `criterion` crate where 500 random elements are inserted into a `SparseArray3` or a `PackedArray` of shapes `[40, 50, 30]` or `[10, 5, 7]`, and it seems like the performances of both are very similar, with `PackedArray` being about ~5-10% faster, depending on the RNG seed

cschwan (approved):
The code is now in a very good shape, I'm going to merge it into `master`.

felixhekhorn:
just for the sake of knowledge transfer: I think I understand the general idea (you're saving coordinates), but `index_mut` could only profit from some more comments :see_no_evil: (as this is clearly a crucial part)

janw20:
Hi Felix, do you mean `///` documentation for the `index_mut` function? Or just in-line comments? Also, how is this normally done, should I commit to this pull request again, and afterward it will be again rebased? Or should I open a new branch on the main repo?

janw20:
I improved the documentation and committed it in https://github.com/NNPDF/pineappl/commit/6a711bbd347a0f1d090ddac043ce0358cc5fa45d to `master`.

PR #280, opened 2024-04-23
Package PineAPPL for Nix

alecandido:
Introduce a Nix package for the PineAPPL CLI.

This should be able to provide a working shell, with PineAPPL and LHAPDF, just by installing Nix and running:
```sh
nix shell github:nnpdf/pineappl nixpkgs#lhapdf
```

(It already works, but you should use `github:nnpdf/pineappl/nix-pkg` to target this branch, and, of course, it will compile PineAPPL on the user machine - to avoid that we should provide a cache)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,25 +26,22 @@
       craneLib = crane.lib.${system};
 
       src = craneLib.cleanCargoSource (craneLib.path ./.);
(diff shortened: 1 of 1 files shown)


alecandido:
This is pretty much ready, and it could be already tested. But I'd wait to merge first #245.

I will further refactor in #281, to make room for the Python package as well.

PR #283, opened 2024-04-29
Add method to change the meaning of particle IDs

cschwan:


pineappl/src/grid.rs
```diff
@@ -2180,6 +2187,33 @@ impl Grid {
...
+            (PidBasis::Evol, PidBasis::Pdg) => {
+                self.lumi = self
+                    .lumi
+                    .iter()
+                    .map(|channel| LumiEntry::translate(channel, &pids::evol_to_pdg_mc_ids))
+                    .collect();
+
+                self.set_key_value("lumi_id_types", "pdg_mc_ids");
+            }
+            (PidBasis::Evol, PidBasis::Evol) | (PidBasis::Pdg, PidBasis::Pdg) => {
```
alecandido, line 2211:
```suggestion
            (x, y) if x == y => {
```

What do you think?

alecandido: Of course you should add `PartialEq` derivation to the struct

cschwan: This doesn't work:

```
error[E0004]: non-exhaustive patterns: `(PidBasis::Pdg, PidBasis::Pdg)` and `(PidBasis::Evol, PidBasis::Evol)` not covered
    --> pineappl/src/grid.rs:2192:15
     |
2192 |         match (self.pid_basis(), pid_basis) {
     |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ patterns `(PidBasis::Pdg, PidBasis::Pdg)` and `(PidBasis::Evol, PidBasis::Evol)` not covered
     |
     = note: the matched value is of type `(PidBasis, PidBasis)`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern, a match arm with multiple or-patterns as shown, or multiple match arms
     |
2211 ~             (x, y) if x == y => {},
2212 ~             (PidBasis::Pdg, PidBasis::Pdg) | (PidBasis::Evol, PidBasis::Evol) => todo!() //(PidBasis::Evol, PidBasis::Evol) | (PidBasis::Pdg, PidBasis::Pdg) => {
     |
```

In other words the compiler isn't clever enough to figure out that these two cases are coverged in the `(x, y) ...` arm.

alecandido: Uhm, the syntax is valid, maybe it is simply not able to check at compile time that it is exhaustive.

You could add another branch:
```rs
    _ => unreachable!();
```
Maybe it's just enough, let me try.

cschwan: Right, but then the original code is simpler and more expressive.

alecandido: I would have said that the original code is more explicit (/verbose...), but the added benefit of the proposal is that it automatically scales to new additions.

However, it was just a proposal, feel free to resolve the conversation.

PR #285, opened 2024-04-30
Generalize the APPLgrid exporter

janw20:
Hi,
I implemented the cases not yet implemented in the APPLgrid exporter. This includes:
- Adding support for exporting PineAPPL grids with multi-dimensional and non-consecutive bins by exporting them as APPLgrids with integer bin limits `0..N`. The ordering is the same as the PineAPPL bins, e.g. given by `pineappl read --bins`. An info message about the bin limits is printed to the user when a PineAPPL grid with multi-dimensional and non-consecutive bins is exported.
- Adding support for arbitrary `Q2`, `x1`, `x2` grids
- Adding the option `epsilon = 1e-12` to `approx_eq!` in lines 240 & 257. Otherwise, this failed for the grid I tested this with, for the two values `0.04979197630496172` and `0.04979197630496281` (the difference is only in the last 3 digits). Or would you handle this differently?

### DIFF
--- a/pineappl_cli/src/export/applgrid.rs
+++ b/pineappl_cli/src/export/applgrid.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, bail, Result};
 use cxx::{let_cxx_string, UniquePtr};
 use float_cmp::approx_eq;
+use itertools::Itertools;
 use ndarray::{s, Axis};
 use pineappl::grid::{Grid, Order};
 use pineappl::subgrid::{Mu2, Subgrid, SubgridParams};
@@ -18,10 +19,10 @@ fn reconstruct_subgrid_params(grid: &Grid, order: usize, bin: usize) -> Result<S
         .subgrids()
         .slice(s![order, bin, ..])
         .iter()
-        .map(|subgrid| {
+        .flat_map(|subgrid| {
             subgrid
                 .mu2_grid()
-                .iter()
+                .into_iter()
                 .map(|&Mu2 { ren, fac }| {
                     if !approx_eq!(f64, ren, fac, ulps = 128) {
                         bail!("subgrid has mur2 != muf2, which APPLgrid does not support");
@@ -29,21 +30,50 @@ fn reconstruct_subgrid_params(grid: &Grid, order: usize, bin: usize) -> Result<S
 
                     Ok(fac)
                 })
-                .collect::<Result<Vec<_>>>()
+                .collect_vec()
         })
-        .collect::<Result<_>>()?;
-    let mut mu2_grid: Vec<_> = mu2_grid.into_iter().flatten().collect();
-    mu2_grid.dedup_by(|a, b| approx_eq!(f64, *a, *b, ulps = 128));
-    let mu2_grid = mu2_grid.as_slice();
+        .collect::<Result<Vec<_>>>()?
+        .into_iter()
+        .sorted_unstable_by(|a, b| a.total_cmp(b))
+        .dedup_by(|&a, &b| approx_eq!(f64, a, b, ulps = 128))
+        .collect_vec();
 
-    if let &[fac] = mu2_grid {
-        result.set_q2_bins(1);
-        result.set_q2_max(fac);
-        result.set_q2_min(fac);
+    let x_grid = grid
+        .subgrids()
+        .slice(s![order, bin, ..])
+        .iter()
+        .flat_map(|subgrid| {
+            [
+                subgrid.x1_grid().into_owned(),
+                subgrid.x2_grid().into_owned(),
+            ]
+            .into_iter()
+        })
+        .flatten()
+        .sorted_unstable_by(|a, b| a.total_cmp(b))
+        .dedup_by(|&a, &b| approx_eq!(f64, a, b, ulps = 128))
+        .collect_vec();
+    
+    result.set_q2_bins(mu2_grid.len());
+    result.set_q2_min(*mu2_grid.first().unwrap());
+    result.set_q2_max(*mu2_grid.last().unwrap());
+    if mu2_grid.len() == 1 {
         result.set_q2_order(0);
+    } else {
+        // TODO: reconstruct the interpolation order. for now leave it as the default
+        // result.set_q2_order(...);
+    }
+    
+    result.set_x_bins(x_grid.len());
+    result.set_x_min(*x_grid.first().unwrap());
+    result.set_x_max(*x_grid.last().unwrap());
+    if x_grid.len() == 1 {
+        result.set_x_order(0);
+    } else {
+        // TODO: reconstruct the interpolation order. for now leave it as the default
+        // result.set_x_order(...);
     }
 
-    // TODO: implement the general case
     Ok(result)
 }
 
@@ -55,15 +85,9 @@ pub fn convert_into_applgrid(
     let bin_info = grid.bin_info();
     let dim = bin_info.dimensions();
 
-    if dim > 1 {
-        bail!(
-            "grid has {} dimensions, but APPLgrid only supports one-dimensional distributions",
-            dim
-        );
-    }
-
-    if bin_info.slices().len() != 1 {
-        bail!("grid has non-consecutive bin limits, which APPLgrid does not support");
+    let integer_bin_limits = dim > 1 || bin_info.slices().len() > 1;
+    if integer_bin_limits {
+        println!("Info: APPLgrid only supports one-dimensional consecutive bin limits, so the bin limits will be set to [0, {}] in steps of 1, corresponding to the ordering given by `pineappl read --bins <GRID>`.", bin_info.bins());
     }
 
     let lumis = grid.lumi().len();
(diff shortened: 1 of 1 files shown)


cschwan:
Concerning 1): this is a good idea, it's surely better to do something than to error out. There's probably no meaningful alternative to the standard bin limits `(0, 1)`, `(1, 2)`, `...` in the general case.

2): The problem with general bin limits is that APPLgrid requires a fixed functional form, which means that the bin limits `x_i` must be the results of a function `x_i = f(i)`, and `f` is hard-coded in APPLgrid. In `LagrangeSubgridV{1,2}` we use the same function and therefore it happens to work in most cases. However, imagine changing the spacing between some of the points between `x_min` and `x_max`; in that case the APPLgrid won't give the right results. In other words: you only ever set `x_min` and `x_max`, but never the points in between, and they can be random.

3) I'd try changing the [`ulps`](https://en.wikipedia.org/wiki/Unit_in_the_last_place) parameter, there shouldn't be such a large difference (try doubling it).

For points 2) and 3) it would be good to have grids to test that with.

janw20:
2. Right, we could throw an error for now if a subgrid is not a `LagrangeSubgridVx`? Also, I did not know how to extract the interpolation order from the grid values, so I am just leaving it at the default value. Is there a way how to reconstruct the interpolation order from just the grid points? Or should the interpolation order maybe be stored in the grid files?
3. At `ulps ≈ 470` this works without the epsilon, below that it still fails for `x = 0.000014007074397454824` and `0.00001400707439745562` in my case. Should I just set it to `512` or is there a more systematic way to determine it? The `float_cmp` author recommends to set both `ulps` and `epsilon`, where `epsilon` is a small multiple of `f64::EPSILON`. For me this starts working with about `5.0 * f64::EPSILON` with `ulps = 128`.

cschwan:
> 2. Right, we could throw an error for now if a subgrid is not a `LagrangeSubgridVx`? Also, I did not know how to extract the interpolation order from the grid values, so I am just leaving it at the default value. Is there a way how to reconstruct the interpolation order from just the grid points? Or should the interpolation order maybe be stored in the grid files?

We can't throw an error, because then optimized grids can't be exported. That would be a big loss.

There's no way to reconstruct the interpolation order from the grid points alone, you can use the 50 that we always use for different interpolation orders. I don't think the interpolation order in the exported APPLgrid is important, usually you don't want to fill them (again).

> 
>     3. At `ulps ≈ 470` this works without the epsilon, below that it still fails for `x = 0.000014007074397454824` and `0.00001400707439745562` in my case. Should I just set it to `512` or is there a more systematic way to determine it? The `float_cmp` author recommends to set both `ulps` and `epsilon`, where `epsilon` is a small multiple of `f64::EPSILON`. For me this starts working with about `5.0 * f64::EPSILON` with `ulps = 128`.

Set it to `512`. I don't think we need an `epsilon` comparison; I believe this is only needed when we compare against a zero result (`0.0 != x` for every non-zero `x` no matter the `ulps` parameter).

cschwan:
> I could also add a `--ulps` argument to `pineappl export`, which lets the user set a different value than the value of `512`, with `512` then being a hard-coded default value?

I don't think this is needed.

janw20:
Okay, I set `ulps = 512` now in the checks of the `x` grid and removed the  `epsilon` argument. From my side this can be merged

cschwan:
What's missing are tests that check the features that you implemented. We need a PineAPPL grid 
1. with multidimensional bin limits that have *bin widths not equal to one*, and one
2. with custom x-grid values.

Those two cases are possibly broken, so we better make sure to test it.

pineappl_cli/src/export/applgrid.rs
```diff
@@ -18,32 +19,61 @@ fn reconstruct_subgrid_params(grid: &Grid, order: usize, bin: usize) -> Result<S
...
-        result.set_q2_min(fac);
+        .collect::<Result<Vec<_>>>()?
+        .into_iter()
+        .sorted_unstable_by(|a, b| a.total_cmp(b))
+        .dedup_by(|&a, &b| approx_eq!(f64, a, b, ulps = 128))
+        .collect_vec();
+
+    let x_grid = grid
+        .subgrids()
+        .slice(s![order, bin, ..])
```
cschwan, line 43: This probably doesn't work quite as expected. APPLgrid stores different channels in a single `appl_igrid`, which has common parameter values `NQ2`, `NQ2min`, `NQ2max`, `NQorder`, `Nx`, `xmin`, `xmax`, `xorder`. This means that:
- if several channels have different `SubgridParam`/`ExtraSubgridParam` values we should error out, unless this is a DIS grid in which we ignore the second dimension. In PineAPPL the second dimension can be either `x1` or `x2`
- the `flat_map` should therefore probably be a simple `map` and we must make sure that for all channels the entries are the same
- the same applies also for `mu_grid`

pineappl_cli/src/export/applgrid.rs
```diff
@@ -100,11 +124,15 @@ pub fn convert_into_applgrid(
     ffi::make_lumi_pdf(id, &combinations).into_raw();
 
     let limits = &bin_info.limits();
-    let limits: Vec<_> = limits
-        .iter()
-        .map(|vec| vec[0].0)
-        .chain(limits.last().map(|vec| vec[0].1))
-        .collect();
+    let limits = if integer_bin_limits {
+        (0..=limits.len()).map(|x| x as f64).collect::<Vec<_>>()
```
cschwan, line 128: Here there's very likely a normalization factor missing. If you change the bin sizes you must correct the cross sections.

pineappl_cli/src/export/applgrid.rs
```diff
@@ -226,12 +254,13 @@ pub fn convert_into_applgrid(
                     .map(|&x2| {
                         appl_x2
                             .iter()
-                            .position(|&x| approx_eq!(f64, x, x2, ulps = 128))
+                            // .position(|&x| approx_eq!(f64, x, x2, ulps = 128, epsilon = 1e-12))
```
cschwan, line 257: Remove the commented out line.

cschwan (changes requested):

pineappl_cli/src/export/applgrid.rs
janw20: The `OPTIMIZE_SUBGRID_TYPE` optimization can change the ends of the `x` and `mu2` grid points, right? So I would check if the grid points are equal modulo some endpoints, and then use the union as the `x` points that will be passed to APPLgrid?

pineappl_cli/src/export/applgrid.rs
janw20: You're right, thanks. Does APPLgrid also divide by the bin widths in the end?

pineappl_cli/src/export/applgrid.rs
cschwan: This would address one point of it, yes.

janw20:
While working on this, I realized: Why are the `SubgridParams` reconstructed? Why aren't the `SubgridParams` from the  `Grid` used, e.g. by exposing them in a `pub fn subgrid_params()`?

cschwan:
> Why are the `SubgridParams` reconstructed? Why aren't the `SubgridParams` from the `Grid` used, e.g. by exposing them in a `pub fn subgrid_params()`?

Not every `Subgrid` necessarily has `SubgridParams` in the way understood by `LagrangeSubgridV{1,2}`. `ImportOnlySubgridV{1,2}` may have arbitrary `x`-node values, for instance linearly-spaced ones like `[0.1, 0.2, 0.3, 0.4, 0.5]`. These points are impossible to generate with the `LagrangeSubgridV{1,2}`.

janw20:
> linearly-spaced ones like `[0.1, 0.2, 0.3, 0.4, 0.5]`
But the exporter can't handle these anyway, since APPLgrid can't handle them, right? (At least that's how I understood your first comment in this pull request.) So it would only matter (for the exporter) how the `LagrangeSubgridV{1,2}` handle the `SubgridParams`. 

Also, is the `Grid::subgrid_params` field updated when optimizing the grid? Because if not, this would even solve the issue of stripped-away grid points at the start and the end of one grid axis.

cschwan:
> > linearly-spaced ones like `[0.1, 0.2, 0.3, 0.4, 0.5]`
> 
> But the exporter can't handle these anyway, since APPLgrid can't handle them, right? (At least that's how I understood your first comment in this pull request.) So it would only matter (for the exporter) how the `LagrangeSubgridV{1,2}` handles the `SubgridParams`.

Yes, APPLgrid only understands the grid-node values spaced as `LagrangeSubgridV{1,2}` uses them.

> 
> The `ExtraSubgridParams` are also not much of an issue, I think, since APPLgrid doesn't support different `x1` and `x2` values.

It does support DIS grids, however, where one of the is trivial (zero and usually one x-grid value in the unused x-dimension)

> 
> Also, is the `Grid::subgrid_params` field updated when optimizing the grid? Because if not, using this field would even solve the issue of stripped-away grid points at the start and the end of the grid axes.

No it's not updated, but you can't rely on this field, because you can merge any kind of subgrid into any Grid. `Grid::subgrid_params` is solely for the purpose of filling temporarily empty subgrids (and this is an optimization, see the implementation of `Grid::fill`).

janw20:
> No it's not updated, but you can't rely on this field, because you can merge any kind of subgrid into any Grid.

I see, this would be a showstopper for using `Grid::subgrid_params` then. Thanks for the clarification.

PR #287, opened 2024-05-10
Introduce more general convolution types

t7phy:
This PR aims to make it possible to optimize grids that perform a double convolution between unpolarized and polarized PDFs or PDFs and FFs by not only checking the PID of the convolution particle but also the convolution type. The grids will require the following new keys:
```
convolution_particle_1: <PID>
convolution_type_1: PDF/FF/polPDF/polFF
convolution_particle_2: <PID>
convolution_type_2: PDF/FF/polPDF/polFF
```


cschwan:
Great @t7phy, thank you very much for this. I will review it as soon as possible!

pineappl/src/grid.rs
```diff
@@ -1389,10 +1389,22 @@ impl Grid {
     }
 
     fn symmetrize_channels(&mut self) {
-        if self.key_values().map_or(false, |map| {
-            map["initial_state_1"] != map["initial_state_2"]
-        }) {
-            return;
+        let map = self.key_values();
+        if map.is_some() {
+            let map = map.unwrap();
```
cschwan, line 1394: Merge this line and the one above into one containing:
```rust
if let Some(map) = map {
    // ...
}
```
This is shorter and you avoid using [`.unwrap()`](https://github.com/NNPDF/pineappl/blob/master/CONTRIBUTING.md#coding-guidelines).

pineappl/src/grid.rs
```diff
@@ -1389,10 +1389,22 @@ impl Grid {
...
+            if map.contains_key("convolution_particle_1")
+                && map.contains_key("convolution_type_1")
+                && map.contains_key("convolution_particle_2")
+                && map.contains_key("convolution_type_2")
+            {
+                if map["convolution_particle_1"] != map["convolution_particle_2"]
+                    || map["convolution_type_1"] != map["convolution_type_2"]
+                {
```
cschwan, line 1402: Similar to the `if`-statement above you can write these two `if`s using `match`:
```rust
match (map.get("convolution_particle_1"), map.get("convolution_particle_2"), map.get("convolution_type_1"), map.get("convolution_type_1")) {
    (Some(convolution_particle_1), Some(convolution_particle_2), Some(convolution_type_1), Some(convolution_type_2)) => (convolution_particle_1 != convolution_particle_2) || (convolution_type_1 != convolution_type_2),
    (None, None, None, None) => {},
    _ => {
        // TODO: if only some of the metadata is set, we should consider this an error
        todo!();
    }
}
```

```

File: pineappl/src/grid.rs [hunk 3]
```diff
@@ -1389,10 +1389,22 @@ impl Grid {
...
+                    return;
+                }
+            } else if map["initial_state_1"] != map["initial_state_2"] {
```
cschwan, line 1405: Make this `else if` an `if`.

pineappl/src/grid.rs
t7phy: done

pineappl/src/grid.rs
t7phy: done, please let me know how to impkement this error

pineappl/src/grid.rs
t7phy: done

cschwan:
I thought about this over the weekend and I think we should refactor this:

1. Introduce a new `enum`:
   ```rust
   enum Convolution {
       UnpolPdf(i32),
       PolPdf(i32),
       UnpolFf(i32),
       // we can add more types as they are needed
   }
   ```
2. Write the method:
   ```rust
   impl Grid {
       pub fn convolutions(&self) -> Vec<Convolution> {
            todo!()
       }
    }
    ```
3. Use the method from 2) in `symmetrize_channels`.

t7phy:
what is the 'convolutions' function supposed to do actually?

cschwan:
> what is the 'convolutions' function supposed to do actually?

It should return a `Vec` with the convolution types for the grid, so for a proton-proton grid we'll return `vec![Pdf(2212), Pdf(2212)]`.

cschwan:
@t7phy I simplified the implementation a bit in commit cc7fd091b01e355b606cf0d16233ac9a45a3333f. What's missing is the following:

- [ ] an exception for the old case: if a metadata `initial_state_1 = id` is set, and the channel has that same `id` set everywhere in the right place, then `Convolution::None` should be returned.
- [ ] remove the methods `Grid::has_pdf1`, `Grid::has_pdf2`, `Grid::initial_state_1` and `Grid::initial_state_2` and use the new method `Grid::convolutions` instead
- [ ] write some small unit tests

cschwan (approved):

pineappl/src/evolution.rs
```diff
@@ -499,10 +499,10 @@ pub(crate) fn evolve_slice_with_one(
         }));
     }
 
-    let pid = if has_pdf1 {
-        grid.initial_state_2()
+    let pid = if grid.convolutions()[0] != Convolution::None {
+        grid.lumi()[0].entry()[0].1
```
felixhekhorn, line 503: this seems wrong to me: lumi and convolutions are playing on two different levels - maybe you meant `grid.convolutions()[1].pid()` here? (same below) (this has the usual caveat mentioned before, that `[1]` has to exist)

pineappl/src/grid.rs
```diff
@@ -222,6 +223,51 @@ pub struct Grid {
     more_members: MoreMembers,
 }
 
+/// Data type that indentifies different types of convolutions.
+#[derive(Debug, Eq, PartialEq)]
```
felixhekhorn, line 227: Do you left `Debug` intentionally?

pineappl/src/grid.rs
```diff
@@ -758,6 +804,93 @@ impl Grid {
...
     }
 
+    /// Return a vector containing the type of convolutions performed with this grid.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the metadata key--value pairs `convolution_particle_1` and `convolution_type_1`,
+    /// or `convolution_particle_2` and `convolution_type_2` are not correctly set.
+    #[must_use]
+    pub fn convolutions(&self) -> Vec<Convolution> {
```
felixhekhorn, line 814: although I suspect the function does a fairly simple job, it looks complicated - could you please add some more (code) comments or break the function into smaller pieces?

pineappl/src/grid.rs
```diff
@@ -758,6 +804,93 @@ impl Grid {
...
+        self.key_values().map_or_else(
+            // if there isn't any metadata, we assume that proton-PDFs are used
```
felixhekhorn, line 816: -> `that two unpolarized proton PDFs are used`

felixhekhorn (commented):
May I add some comments here?

pineappl/src/grid.rs
cschwan: `Debug` is needed whenever you compare objects using assertions, which is done here:

https://github.com/NNPDF/pineappl/blob/5eba2360f84c503e2a1fc4b1c8c9f49757cd5f93/pineappl/src/grid.rs#L1925-L1951

pineappl/src/grid.rs
cschwan: I added more comments in commit ba94ab9cf1a300b7727c57ed4d05a87b86788f93 - this function will become a simple getter once we save the convolution types as members of a `Grid` using a new file format.

pineappl/src/grid.rs
cschwan: Changed in commit ba94ab9cf1a300b7727c57ed4d05a87b86788f93.

Issue #288, opened 2024-05-30
Extend the evolution to accommodate for two different EKOs


Radonirinaunimi:
Falls in https://github.com/NNPDF/pineappl/issues/172 and necessary for https://github.com/NNPDF/pineko/issues/180. 

Given that now different initial states can be dealt with in pineappl during the convolution, the evolution part needs also to be extended to accommodate for two different (un-/polarized time-/space-like) EKOs.

PR #289, opened 2024-05-30
Accommodate for two different EKOs in Evolution

Radonirinaunimi:
Addresses https://github.com/NNPDF/pineappl/issues/288. So far, this is just a bare brute-force implementation to test if it conceptually works.

Remaining to do:
- [x] add some unit tests
- [x] ~add checks that verify the compatibility between the initial states and the EKOs~ (Moved to Pineko)
- [x] Fix the CLI to accept different EKOs
- [x] deprecate `Grid::evolve` in favor of `Grid::evolve_with_slice_iter`
- [x] simplify the passing of EKOs
- [x] make sure changes from #293 are taken into account: before merging this branch into master
  - [x] after having merged #293 into master
  - [x] merge master into this branch
- [ ] remove the methods `Grid::evolve2` and its Python wrapper; confirm that you don't use it in Pineko


pineappl/src/grid.rs
```diff
@@ -1384,6 +1384,8 @@ impl Grid {
         }
     }
 
+    /// TODO: Fix to account for the different EKOs
+    /// TODO: Does the info have to be different?
```
felixhekhorn, line 1388: yes, they will be different - there has to be one info for one EKO; e.g. there are typically less polarized flavor than unpolarized

Radonirinaunimi: So, practically, what would be different in each case from the following attributes?
https://github.com/NNPDF/pineappl/blob/2a8a9912a29d81d8b4ca4906eeec8daf6f5f80ea/pineappl_py/src/grid.rs#L474-L486

cschwan:
This goes in the right direction, but I'd like to generalize this even further for the cases where we have 3 or even arbitrarily many evolutions.

Now that we have `Grid::convolutions`, we can use it to check which convolutions require different EKOs. Then we'll need - I suppose - a single object that returns EKOs for each `Convolution` type. This object should replace `operator_a`, `operator_b` and `slice_info`.

felixhekhorn:
I wonder if this feature is a priority? since fitting with several distributions is rather a mid- or long-term feature ... in the short term we want to close one convolution (with a fixed PDF), produce a new grid with only one open convolution and turn that into an FK table, wouldn't you say so?

Radonirinaunimi:
> I wonder if this feature is a priority? since fitting with several distributions is rather a mid- or long-term feature ... in the short term we want to close one convolution (with a fixed PDF), produce a new grid with only one open convolution and turn that into an FK table, wouldn't you say so?

That's absolutely true! But if this could be achieved with more or less the same efforts, that'll be a better no? But I agree that if this turns out to be more difficult, we could delay it for later.

cschwan:
I believe a priority should be to have a stable interface, which can be stable for a long time. This means it should account for all the various EKOs possible, and then the most general *implementation* can take longer as I already said in <https://github.com/NNPDF/pineappl/pull/289#issuecomment-2141576405>.

Note that for now, we don't even use the new `Grid::evolve_with_slice_iter` in Pineko, which already is the third iteration of evolution (first with `Grid::convolute_eko`, second was `Grid::evolve`).

Radonirinaunimi:
> Now that we have Grid::convolutions, we can use it to check which convolutions require different EKOs.

Thinking about this a bit, we can use `Grid::convolutions` to check if the grid requires different EKOs, but ultimately the different EKOs have to be passed as inputs from pineko - and this is really what should define the "EKO" object. 

So, I think the easiest and best solution would be to pass a list of EKOs to:
https://github.com/NNPDF/pineappl/blob/95b3bc36c477d379744b0dd1ad834d45b10e15b0/pineappl/src/grid.rs#L1400-L1406

in that we replace `operator_` with `list_operators`, and then check with `Grid::convolutions` that the number and types of EKOs is consistent with the initial states.

In this way, we can arbitrarily pass different evolutions. Does this sound reasonable?

felixhekhorn:
@cschwan original idea was based on an iterative approach: calling `evolve` several times with one operator each and closing one (multi-)index at a time - I think this might be conceptually easier then dealing with all convolutions at the same time. Ofc there can always be some sugar to do all in one go

cschwan:
> @cschwan original idea was based on an iterative approach: calling `evolve` several times with one operator each and closing one (multi-)index at a time - I think this might be conceptually easier then dealing with all convolutions at the same time. Ofc there can always be some sugar to do all in one go

Indeed, but using this approach it's easy to get it wrong, so I think I've changed my mind :smile:.

cschwan:
> In this way, we can arbitrarily pass different evolutions. Does this sound reasonable?

That's what I had in mind, although I'm still unclear about the details.

giacomomagni:
Hi all,
since now we made quite a few progress in https://github.com/NNPDF/pineko/pull/181 and we are ready to support
`pineappl==0.7.4` in Pineko, would it be possible to merge this PR and have a new micro release, so these changes 
are more accessible. 
In particular we will need to introduce `FkTable.convolve_with_two` also in `n3fit`. 
Thanks!!

Radonirinaunimi:
@cschwan, @felixhekhorn: After having looked a bit, I think generalizing to any arbitrary evolution is complicated and requires a lot of changes (because for example, most of the structs will have to be modified).

As far as this PR is concerned, this already achieves the initial intended purposes, and we already tested it in https://github.com/NNPDF/pineko/pull/181. In this sense, at least, I think this can be reviewed.

pineappl_cli/src/evolve.rs
```diff
@@ -425,6 +425,7 @@ mod eko {
 fn evolve_grid(
     grid: &Grid,
     eko: &Path,
+    extra_eko: &Path,
```
cschwan: Here you can already start generalizing the code: replace `eko` and `extra_eko` with an `ekos: &[&Path]`. You can then use `match ekos.len() { 0 => /* Error */, 1 | 2 => {}, _ => unimplemented!() }` to catch the cases that don't make sense or aren't implemented yet.

pineappl/src/grid.rs
```diff
@@ -1446,16 +1466,21 @@ impl Grid {
     pub fn evolve_with_slice_iter<'a, E: Into<anyhow::Error>>(
```
cschwan, line 1475: I'm uncertain whether always passing two EKOs is a good idea. Note that this is a performance-critical section part of PineAPPL, and we don't want to unnecessarily slow down FK-table generation. Some jet grids take up to five hours if I remember correctly. We must check the impact of this patch on the performance.

pineappl/src/evolution.rs
```diff
@@ -552,11 +556,11 @@ pub(crate) fn evolve_slice_with_two(
     let mut operators_b = Vec::new();
 
     for subgrids_ol in grid.subgrids().axis_iter(Axis(1)) {
-        let mut tables = vec![Array2::zeros((info.x0.len(), info.x0.len())); channels0.len()];
+        let mut tables = vec![Array2::zeros((info_a.x0.len(), info_a.x0.len())); channels0.len()];
```
cschwan, line 559: Shouldn't this be `info_b` in the second case?

cschwan (changes requested):
We also need an example to add an integration test; ideally a simple-enough grid with two EKOs. If the two EKOs have differing `x_grid` that would be even better, to properly test whether the correct `info_a` or `info_b` structure was used.

pineappl/src/grid.rs
```diff
@@ -1397,14 +1399,32 @@ impl Grid {
     #[deprecated(since = "0.7.4", note = "use evolve_with_slice_iter instead")]
     pub fn evolve(
```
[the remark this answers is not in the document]
Radonirinaunimi: Thanks a lot @cschwan for the comments!

Re this comment, at the moment the python interface relies on `Grid::evolve`, do you suggest to modify it to rely on `Grid::evolve_with_slice_iter` instead?

pineappl/src/evolution.rs
```diff
@@ -522,26 +522,30 @@ pub(crate) fn evolve_slice_with_one(
...
-    })?;
+    let (pid_indices_a, pids_a) =
+        pid_slices(input_operator_a, info_a, gluon_has_pid_zero, &|pid1| {
+            grid.channels()
+                .iter()
+                .flat_map(Channel::entry)
+                .any(|&(a, _, _)| a == pid1)
+        })?;
+    let (pid_indices_b, pids_b) =
+        pid_slices(input_operator_b, info_a, gluon_has_pid_zero, &|pid1| {
```
felixhekhorn, line 543: is the `info_a` intentional?

pineappl/src/grid.rs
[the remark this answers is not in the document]
cschwan: Yes, although this is probably not necessarily connected to this PR, but was rather a task for <https://github.com/NNPDF/pineappl/pull/244>. I suggest you write a new method, `evolve_with_two` for instance, and then use that.

pineappl/src/evolution.rs
Radonirinaunimi: If think here using `info_a` or `info_b` is exactly the same. But just to be always certain, I can add a check that this is always the case, and otherwise panic.

pineappl/src/grid.rs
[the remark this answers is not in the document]
Radonirinaunimi: > Yes, although this is probably not necessarily connected to this PR, but was rather a task for https://github.com/NNPDF/pineappl/pull/244. I suggest you write a new method, evolve_with_two for instance, and then use that.

Ah, ok! Yes, I can take care of doing so in this PR:

Radonirinaunimi:
So the status is that with the modifications in https://github.com/NNPDF/pineko/pull/181, this works fine. Below, for instance, is a comparison between the grids (with PDF evolution) and the FK tables (with EKO) for the observable $\Delta \sigma (p_{\mathrm{pol}} \otimes p_{\mathrm{unpol}})$.

```bash
*********************************
Th: 823  STAR_WMWP_510GEV_WM-AL-POL
*********************************
        FKTable          Grid  abs_diff  rel_diff
0    803.274230    803.188890  0.085339  0.010625
1   5864.619611   5864.785764 -0.166153 -0.002833
2  14206.376784  14206.816815 -0.440031 -0.003097
3  20814.990536  20815.514610 -0.524074 -0.002518
4  19516.276622  19517.356552 -1.079930 -0.005533
5   7529.690834   7531.065920 -1.375087 -0.018259
```
Will now address all the comments above and add tests.

pineappl/src/grid.rs
Radonirinaunimi: @giacomomagni said that the longest jet now takes about 30mn. But I absolutely agree that we should thoroughly test how much does this affect the performance, and if it does in some significant way implement something better.

cschwan:
> Will now address all the comments above and add tests.

You can add a test similar to other tests, for instance like:

https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/pineappl_cli/tests/evolve.rs#L429-L446

Simply create another sensibly-named test function, add a new static string with the test output and set the argument of `stderr` to `""`.

For that to work in the CI,
- upload both operators and the grid to <https://data.nnpdf.science/pineappl/test-data/>,
- add another `curl` invocation in the Rust action here: https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/.github/workflows/rust.yml#L32 and also in the bash script: https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/maintainer/generate-coverage.sh#L7 with `wget`, and finally
- increment the test-data version number here: https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/.github/workflows/rust.yml#L26

Radonirinaunimi:
Thanks @cschwan! I will follow this.

felixhekhorn:
> > Will now address all the comments above and add tests.
> 
> You can add a test similar to other tests, for instance like:
> 
> https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/pineappl_cli/tests/evolve.rs#L429-L446
> 
> Simply create another sensibly-named test function, add a new static string with the test output and set the argument of `stderr` to `""`.
> 
> For that to work in the CI,
> 
>     * upload both operators and the grid to https://data.nnpdf.science/pineappl/test-data/,
> 
>     * add another `curl` invocation in the Rust action here: https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/.github/workflows/rust.yml#L32
>        and also in the bash script: https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/maintainer/generate-coverage.sh#L7
>        with `wget`, and finally
> 
>     * increment the test-data version number here: https://github.com/NNPDF/pineappl/blob/cca611189eb420bb0bfe26b96e3b6bbdec0c5a11/.github/workflows/rust.yml#L26

@cschwan I wonder if you should add these instructions to https://github.com/NNPDF/pineappl/blob/master/maintainer/README.md ?

cschwan:
Yes, that's a good idea; I'll add them to `CONTRIBUTING.md`, since every contributor can potentially add these kind of tests.

Radonirinaunimi:
The test is failing because the container registry needs to be updated. I am not sure who has the permission to do so? Maybe we should also add something in this regard in the `CONTRIBUTING.md`?

cschwan:
I added the instructions with commit 644677b858615f22afb1fbda2dc00518b818e6c3. I suggest first try fix the problems without updating the container, because this usually isn't as straightforward as it should be. @Radonirinaunimi please fix the test that's not related to fetching the PDFs, I will try to fix the other problems.

cschwan:
@Radonirinaunimi in commit 5630f708fc96138afee785eba3f5a57caa5ac292 you've changed the implementation of the convolution function of the CLI, which heavily overlaps with what I'm doing with @t7phy in #293. For the being I'm OK with leaving it as is, but we should make sure that we first merge #293 into master, master into this branch and then this branch into master. I'll add a TODO item for this above.

Radonirinaunimi:
> I added the instructions with commit [644677b](https://github.com/NNPDF/pineappl/commit/644677b858615f22afb1fbda2dc00518b818e6c3). 

Perfect, thanks!

> I suggest first try fix the problems without updating the container, because this usually isn't as straightforward as it should be. @Radonirinaunimi please fix the test that's not related to fetching the PDFs, I will try to fix the other problems.

Ok, I think I will just move the line here:
https://github.com/NNPDF/pineappl/blob/5630f708fc96138afee785eba3f5a57caa5ac292/maintainer/pineappl-ci/script.sh#L55-L56
to [rust.yml](https://github.com/NNPDF/pineappl/blob/extend_eko_convolution/.github/workflows/rust.yml) for the time being then. For the test to pass, we really need a PDF set evolved with the same EKO.

> @Radonirinaunimi in commit https://github.com/NNPDF/pineappl/commit/5630f708fc96138afee785eba3f5a57caa5ac292 you've changed the implementation of the convolution function of the CLI, which heavily overlaps with what I'm doing with @t7phy in https://github.com/NNPDF/pineappl/pull/293. For the being I'm OK with leaving it as is, but we should make sure that we first merge https://github.com/NNPDF/pineappl/pull/293 into master, master into this branch and then this branch into master. I'll add a TODO item for this above.

Yes, I was indeed aware of this and I extended the CLI just for the sake of running tests but having in mind that this will be overwritten. We simply need to revert some of the changes in [5630f70](https://github.com/NNPDF/pineappl/pull/289/commits/5630f708fc96138afee785eba3f5a57caa5ac292). So the plans and todos sound good. Just for info, about when do you expect https://github.com/NNPDF/pineappl/pull/293 to get merged?

cschwan:
> > I added the instructions with commit [644677b](https://github.com/NNPDF/pineappl/commit/644677b858615f22afb1fbda2dc00518b818e6c3).
> 
> Perfect, thanks!
> 
> > I suggest first try fix the problems without updating the container, because this usually isn't as straightforward as it should be. @Radonirinaunimi please fix the test that's not related to fetching the PDFs, I will try to fix the other problems.
> 
> Ok, I think I will just move the line here:
> 
> https://github.com/NNPDF/pineappl/blob/5630f708fc96138afee785eba3f5a57caa5ac292/maintainer/pineappl-ci/script.sh#L55-L56
> 
> to [rust.yml](https://github.com/NNPDF/pineappl/blob/extend_eko_convolution/.github/workflows/rust.yml) for the time being then. For the test to pass, we really need a PDF set evolved with the same EKO.

Right, that's even simpler than what I thought of; please go ahead with this.

> 
> > @Radonirinaunimi in commit [5630f70](https://github.com/NNPDF/pineappl/commit/5630f708fc96138afee785eba3f5a57caa5ac292) you've changed the implementation of the convolution function of the CLI, which heavily overlaps with what I'm doing with @t7phy in #293. For the being I'm OK with leaving it as is, but we should make sure that we first merge #293 into master, master into this branch and then this branch into master. I'll add a TODO item for this above.
> 
> Yes, I was indeed aware of this and I extended the CLI just for the sake of running tests but having in mind that this will be overwritten. We simply need to revert some of the changes in [5630f70](https://github.com/NNPDF/pineappl/pull/289/commits/5630f708fc96138afee785eba3f5a57caa5ac292). So the plans and todos sound good. Just for info, about when do you expect #293 to get merged?

Pretty soon I hope.

cschwan:
@Radonirinaunimi the newest container now has the polarized PDF included, if you like you can test that by removing the PDF set from the action.

Radonirinaunimi:
> @Radonirinaunimi the newest container now has the polarized PDF included, if you like you can test that by removing the PDF set from the action.

Thanks a lot @cschwan! I've removed the download of the pPDF set from `rust.yml`.

cschwan:
At some point the pipeline will probably break again, due to the forced migration from Node 16 to 20, see here: <https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/>. In that case let's cherry-pick this commit: 0a8ea2a2737e1318f46c0101e0521d5661aa6193.

@Radonirinaunimi, @t7phy: In master I pushed commit ed8601d97fce4a21e48d1a5f427e44891c89efe9 that prepares the `evolve_slice_with_two` function for `n >= 2` convolutions. This isn't exactly the scope of this pull request, but maybe you'd like to have a look at it. I checked that the performance is unaffected from this patch for `ATLAS_1JET_8TEV_R06`.

Finally, I managed to find out how to pass Iterators from Python to Rust. @giacomomagni @Radonirinaunimi we need to agree on an interface that supports multiple EKOs and convolutions. From Pineko we need a list of tuples,

    [(Convolution, Iterator), ...]

The iterators must return a tuple of a four-dimensional EKO slice together an object of type `OperatorSliceInfo`:

    Iterator -> (OperatorSliceInfo, ndarray4)

Do you think this can be easily done?

giacomomagni:
> The iterators must return a tuple of a four-dimensional EKO slice together an object of type `OperatorSliceInfo`:
> ```
> Iterator -> (OperatorSliceInfo, ndarray4)
> ```
> Do you think this can be easily done?

Thanks a lot @cschwan. Yes, this is doable inside pineko. Once we have a working version here I will update https://github.com/NNPDF/pineko/pull/181 accordingly.

Radonirinaunimi:
Thanks a lot @cschwan!

> At some point the pipeline will probably break again, due to the forced migration from Node 16 to 20, see here: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/. In that case let's cherry-pick this commit: [0a8ea2a](https://github.com/NNPDF/pineappl/commit/0a8ea2a2737e1318f46c0101e0521d5661aa6193).
> 
I think this should be fine. We can introduce all the (so far) breaking changes in one release and then propagate throughout the rest of the pipeline.

> @Radonirinaunimi, @t7phy: In master I pushed commit [ed8601d](https://github.com/NNPDF/pineappl/commit/ed8601d97fce4a21e48d1a5f427e44891c89efe9) that prepares the `evolve_slice_with_two` function for `n >= 2` convolutions. This isn't exactly the scope of this pull request, but maybe you'd like to have a look at it. I checked that the performance is unaffected from this patch for `ATLAS_1JET_8TEV_R06`.
> 
I will have a look asap and reflect the changes in here.

> Finally, I managed to find out how to pass Iterators from Python to Rust. @giacomomagni @Radonirinaunimi we need to agree on an interface that supports multiple EKOs and convolutions. From Pineko we need a list of tuples,
> 
> ```
> [(Convolution, Iterator), ...]
> ```
> 
> The iterators must return a tuple of a four-dimensional EKO slice together an object of type `OperatorSliceInfo`:
> 
> ```
> Iterator -> (OperatorSliceInfo, ndarray4)
> ```
> 
> Do you think this can be easily done?

Yes, this is indeed perfectly feasible and in terms of interface arguably the best solution to finally get rid of `evolve` while at same time allowing for any number of ekos to be passed.

cschwan:
At some point we'll have to change the interface slightly again, because different convolutions will be associated with different scales (PDFs with the factorization scale and FFs with the fragmentation scale). However I can't exactly foresee what will change, so we shouldn't worry about it in this PR.

giacomomagni:
> At some point we'll have to change the interface slightly again, because different convolutions will be associated with different scales (PDFs with the factorization scale and FFs with the fragmentation scale). However I can't exactly foresee what will change, so we shouldn't worry about it in this PR.

I guess the solution is to add a `Q2` variable to the `tuple`, but right now we don't even have this scale in the nnpdf theories, so I agree to postpone the change.

pineappl/src/grid.rs
cschwan: This shouldn't have an impact on the runtime since we have two separate operators in any case. If we pass the same operator twice we may waste some memory, but that's a problem we can address later.

pineappl/src/grid.rs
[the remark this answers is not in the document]
cschwan: I addressed this in commit 6978daa86477cd2c20597b08c5f88f6fe808ad0a: the new methods are now named similarly as the old ones, but they have the suffix `2`, so the method `evolve` stays the same and the new method is called `evolve2`. Make sure to propagate the changes into Pineko.

pineappl_py/pineappl/grid.py
```diff
@@ -298,6 +298,66 @@ def convolve_with_two(
...
+                np.array(operators["targetgrid"]),
+                np.array(mur2_grid, dtype=np.float64),
+                np.array(alphas_values, dtype=np.float64),
+                xi,
+                lumi_id_types,
+                np.array(order_mask, dtype=bool),
+            )
+        )
+
+    def evolve2(
```
felixhekhorn, line 360: I guess you want to prepend the `2` also to the Python call below

cschwan: Yes we want that, good catch!

cschwan: Fixed in commit 6081b979ff2b0448797ced6a5beedd267d8ceb6a.

pineappl/src/evolution.rs
cschwan: I believe it's possible that different hadronic initial states have different `x` grids, which then would be reflected in different `infos`. I fixed most of that in commit 4a8f7a492118aea8abf74650c9e0757f6027c895.

cschwan:
@Radonirinaunimi I've implemented most of the necessary changes, in particular the Python binding for `Grid::evolve_with_slice_iter2`. In `pineappl_py/tests/test_fk_table.py` you can see an example how to use this function. Could you try to adapt Pineko to use this function?

giacomomagni:
> @Radonirinaunimi I've implemented most of the necessary changes, in particular the Python binding for `Grid::evolve_with_slice_iter2`. In `pineappl_py/tests/test_fk_table.py` you can see an example how to use this function. Could you try to adapt Pineko to use this function?

This seems to work https://github.com/NNPDF/pineko/pull/181/commits/d62dac4e332529f016f7e139cad28cf17bfd77a8 
thanks @cschwan

Radonirinaunimi:
I finally had time to carefully look into this again. Thanks again the for changes @cschwan (and especially for the improvements), these look really good to me. I believe we've achieved what we ought to do in this PR, should this be merged now?

cschwan:
We didn't manage to fully generalize the evolution interfaces, but that wasn't the intention of this pull request so that's OK. In the future, we'll have to change it again, but so is life.

@Radonirinaunimi we can almost merge, can you please take care of the following:

- [ ] remove the methods `Grid::evolve2` and its Python wrapper; confirm that you don't use it in Pineko

Radonirinaunimi:
> We didn't manage to fully generalize the evolution interfaces, but that wasn't the intention of this pull request so that's OK. In the future, we'll have to change it again, but so is life.
> 
> @Radonirinaunimi we can almost merge, can you please take care of the following:
> 
>     * [ ]  remove the methods `Grid::evolve2` and its Python wrapper; confirm that you don't use it in Pineko

Yes, I can do this.

cschwan:
When you've removed the methods, you can merge this branch into `master`. Then I'm going to prepare a new release `0.8.1` that will have the new method.

Radonirinaunimi:
> When you've removed the methods, you can merge this branch into `master`. Then I'm going to prepare a new release `0.8.1` that will have the new method.

Perfect! Awesome.

The method has been removed, I will wait for the tests to succeed and then merge. Once `v0.8.1` is available I will propagate it to `nnpdf` and `pineko` to make the CI working again.

Thanks again 🚀

cschwan:
Great, thank you!

cschwan (approved):

PR #292, opened 2024-06-06
Use `nano-gemm` instead of `matrixmultiply`

cschwan:
Here are some benchmarks of the existing code:

- <https://github.com/NNPDF/pineappl/pull/244#issuecomment-1762853072>
- <https://github.com/NNPDF/pineappl/pull/244#issuecomment-1772808575>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -469,6 +469,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "equator"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c35da53b5a021d2484a7cc49b2ac7f2d840f8236a286f84202369bd338d761ea"
+dependencies = [
+ "equator-macro",
+]
+
+[[package]]
+name = "equator-macro"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3bf679796c0322556351f287a51b49e48f7c4986e727b5dd78c972d30e2e16cc"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "equivalent"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


cschwan:
Using the 22 GB-sized flavour-basis EKO `ATLAS_1JET_8TEV_R06` for a strong coupling of `0.119` at the Z-boson mass, the old code takes
```
real    160m21.133s
user    159m53.132s
sys     0m27.700s
```
to evolve the corresponding grid. With `nano-gemm` it only takes
```
real    121m36.539s
user    121m9.532s
sys     0m26.784s
```
That's a wonderful **25% reduction in runtime**!

BTW: the old runtime (with OpenBLAS even) was
```
real    301m45.642s
user    300m24.360s
sys     0m54.963s
```
and the difference comes from the fact that I used an evolution-basis EKO that's much bigger due to the rotation: 45 GB.

mert-kurttutan:
Hi,
Your problem seems interesting and practical. Could please you share the steps to reproduce your results (along with the specs of the hardware you used) if possible?
I might be able to contribute.

cschwan:
Hi @mert-kurttutan,

The linear algebra routines are used in an operation that we call 'evolution', and some faster running evolutions are used in our integration tests. See this file: [`pineappl_cli/tests/evolve.rs`](https://github.com/NNPDF/pineappl/blob/master/pineappl_cli/tests/evolve.rs). These tests run the binary that we call the 'PineAPPL CLI', and for these cases we always run
```shell
pineappl evolve <INPUT> <EKO> <OUTPUT> <CONV_FUNS>
```
The integration tests simply verify the output.

For examples of how the arguments are used have a look at the tests. To run them successfully, you'll need the test data that is passed to the CLI, which you can download [here](https://data.nnpdf.science/pineappl/test-data). It's probably easier to copy and run the `wget` calls from [`maintainer/generate-coverage.sh`](https://github.com/NNPDF/pineappl/blob/master/maintainer/generate-coverage.sh). The files must be placed into a folder `test-data` at top level of the repository.

The installation is probably a bit tricky, please read <https://nnpdf.github.io/pineappl/docs/installation.html#cli-pineappl-for-your-shell> for instructions (you will need the evolution feature, all other optional features are not needed). However, before compiling the Rust code, you'll need to install LHAPDF 6.5.4; without this C++ library nothing will compile/run unfortunately. The installation instructions for this library are [here](https://lhapdf.hepforge.org/install.html).

If you have suggestion on how to improve these documents we'd be happy to take your comments into account (best in a separate Issue). While I'm writing this, I realized that in our installation documents we never mention LHAPDF, probably because practically everyone in our community has it installed. This has to be improved.

cschwan:
Starting from commit 9ca3022ef8288b9360584ced5d78f6701f7667e9 the new baseline is now:

```
real    4m54.425s
user    4m20.387s
sys     0m32.138s
```

Apparently we did a lot of linear algebra with zeros :disappointed:.

cschwan:
Since commit 9ca3022ef8288b9360584ced5d78f6701f7667e9 the performance is acceptable and I'd like to avoid adding more dependencies, so I'm closing this as WONTFIX.

@mert-kurttutan if you're still interested and/or have question please let me know!

PR #293, opened 2024-06-10
Allow to specify multiple convolution functions with CLI

### LINKED ISSUE #174: Add support for different initial-state hadrons in the CLI

cschwan:
This will close #174.

TODO:
- [x] migrate subcommand `analyze`; done in commit cfcf02711cebb49028327012a4d31f9a7dfabecc
- [x] migrate subcommand `convolve`; done in commit 14366184659adeaaac6930abb127e24a9483499a
- [x] migrate subcommand `plot`; done in commit 1d4af9614b9338d477ab56a673b31a2685bc1940 and ddbf227937af035574372d6a76a34c990dc71d93
- [x] migrate subcommand `pull`; done in commit a29639bd476971710672ec83fd6239832f532c9e
- [x] migrate subcommand `uncert`; done in commit 5f8e576bff7ef0ea31251ea034de8d76bf430c2a


pineappl_cli/src/helpers.rs
```diff
@@ -207,11 +207,55 @@ pub fn convolve_scales(
...
+            let x_min2 = fun2.x_min();
+            // is the following line needed?
+            // let mut alphas2 = |q2| fun2.alphas_q2(q2);
+            let mut fun2 = |id, x, q2| {
+                if !cfg.allow_extrapolation && (x < x_min2 || x > x_max2) {
+                    0.0
+                } else {
+                    fun2.xfx_q2(id, x, q2)
+                }
+            };
```
felixhekhorn, line 252: can this be abstracted into a function? it is now repeated for the third time ...

cschwan: I just tried that, but it is difficult to achieve, because you need a function that constructs a callable, and then it get non-trivial to satisfy the borrow checker...

felixhekhorn: Mmm ... can't you do something like
```rust
let mut fun2 = |id,x,q2| { masked_xfx_q2(fun2,cfg.allow_extrapolation, id,x,q2) }
```
the you don't need to return a function but can evaluate it ...

cschwan: The way you wrote it I'd have to call `Pdf::x_max` and `Pdf::x_min` in `masked_xfx_q2`, which is not a good idea.

felixhekhorn: why? (I'm happy to pay 1% speed for the code reduction)

cschwan:
While converting some of the code to use the new `ConvFun` structs I stumbled upon a new question we need answer: how do we calculate PDF/FF uncertainties if two PDFs are given, for instance?

cschwan:
> While converting some of the code to use the new `ConvFun` structs I stumbled upon a new question we need answer: how do we calculate PDF/FF uncertainties if two PDFs are given, for instance?

To answer my question, it is instructive to think about the following situation: one PDF, one FF. In that case we now have two uncertainties: a PDF uncertainty and a FF uncertainty. In general we thus have an uncertainty associated to each one of the different convolution functions. We can calculate the uncertainty by varying the members of this set, while keeping the members of the other sets fixed (to the central member, for instance).

felixhekhorn:
> > While converting some of the code to use the new `ConvFun` structs I stumbled upon a new question we need answer: how do we calculate PDF/FF uncertainties if two PDFs are given, for instance?
> 
> To answer my question, it is instructive to think about the following situation: one PDF, one FF. In that case we now have two uncertainties: a PDF uncertainty and a FF uncertainty. In general we thus have an uncertainty associated to each one of the different convolution functions. We can calculate the uncertainty by varying the members of this set, while keeping the members of the other sets fixed (to the central member, for instance).

I think for the moment this is what needs to be done - but I just wanted to point out that this might (or might not) be more complicated in principle: this assumes that the two ($N$) convolutions are uncorrelated (which for now they are). However, if one day we decide to fit PDF and FF _simultaneously_ this assumption would fail and you would need to vary them consistently.

cschwan:
> [..] I just wanted to point out that this might (or might not) be more complicated in principle: this assumes that the two (N) convolutions are uncorrelated (which for now they are). However, if one day we decide to fit PDF and FF _simultaneously_ this assumption would fail and you would need to vary them consistently.

Ah, that's a good point, thanks for pointing that out. In that case we will very likely need support from a convolution function library ('partons') that will tell us how exactly the uncertainty has to be calculated.

pineappl_cli/tests/uncert.rs
```diff
@@ -24,56 +24,57 @@ Options:
...
-1 2.25  2.5 6.9028342e2 6.9027941e2 -1.16 1.16
-2  2.5 2.75 6.0025198e2 6.0022595e2 -1.18 1.18
-3 2.75    3 4.8552235e2 4.8548211e2 -1.22 1.22
-4    3 3.25 3.6195456e2 3.6191001e2 -1.27 1.27
-5 3.25  3.5 2.4586691e2 2.4582640e2 -1.35 1.35
-6  3.5    4 1.1586851e2 1.1584074e2 -1.51 1.51
-7    4  4.5 2.7517266e1 2.7504644e1 -2.77 2.77
+const DEFAULT_STR: &str =
+    "b   etal    dsig/detal    NNPDF31_nlo_as_0118_luxqed       NNPDF40_nnlo_as_01180   
+     []        [pb]
```
felixhekhorn, line 29: wouldn't it be better to have ` [] [pb] [pb] [%] [%] [pb] [%] [%]` ? such that it is more plain that the 6th column is a _relative_ uncert? (here, by chance, it is almost clear) (also just one `[%]` as before would work)

cschwan: Fixed in commit a981f7e2241410c1c73bc495aebdda3c7a64e306.

The problem is that the [`prettytable`](https://crates.io/crates/prettytable-rs) crate supports only a single title row. We therefore use a linebreak to simulate a second row, but in this particular case we use a single title for three columns, and together with the linebreak this effectively disables the correct centering of the units. However, I agree that even with the wrong formatting having the units is important.

PR #295, opened 2024-06-29
Make the CLI and extra dependency of the Python package

alecandido:
This is a tiny consequence of #266, just to simplify the installation.

It will cause an overhead, but I'd assume that most people running
```sh
pip install pineappl
```
would like to have the CLI as well.

### DIFF
--- a/pineappl_py/pyproject.toml
+++ b/pineappl_py/pyproject.toml
@@ -21,10 +21,15 @@ classifiers = [
   "Topic :: Scientific/Engineering :: Physics",
 ]
 
-dependencies = ["numpy>=1.16.0,<2.0.0"]
+dependencies = ["numpy>=1.16.0,<2.0.0", "pineappl-cli"]
 
 [project.optional-dependencies]
-docs = ["sphinx>=6.2.1", "sphinx_rtd_theme>=1.2.2", "sphinxcontrib-bibtex>=2.5.0", "nbsphinx>=0.9.2"]
+docs = [
+  "sphinx>=6.2.1",
+  "sphinx_rtd_theme>=1.2.2",
+  "sphinxcontrib-bibtex>=2.5.0",
+  "nbsphinx>=0.9.2",
+]
 test = ["pytest", "pytest-cov"]
 
 [project.urls]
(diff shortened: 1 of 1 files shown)


cschwan:
I don't think this is a good idea:
1. the Python interface doesn't depend on the CLI in the strict sense
2. on Windows this will lead to an error because there's a Python interface, but no CLI.

It's not a dependency, but rather a recommended/suggested package. I don't know if Python has a concept for those.

alecandido:
> It's not a dependency, but rather a recommended/suggested package. I don't know if Python has a concept for those.

There are *extras*, such that you can install it as:
```sh
pip install pineappl[cli]
```

> the Python interface doesn't depend on the CLI in the strict sense

True, but it's convenient to install them together. The main motivation is for dependent programs, not for end-users.
Unfortunately, the missing bit in Python, which would be convenient in this case, are *default extras* (default features in Rust, as it is often `std`).

> on Windows this will lead to an error because there's a Python interface, but no CLI.

This would not be a problem. Dependencies can be specified as platform-dependent, and we could exclude Windows (until there will be a released wheel for that as well).

cschwan:
> > It's not a dependency, but rather a recommended/suggested package. I don't know if Python has a concept for those.
> 
> There are _extras_, such that you can install it as:
> 
> ```shell
> pip install pineappl[cli]
> ```

Aah yes, that's perfect!

> > the Python interface doesn't depend on the CLI in the strict sense
> 
> True, but it's convenient to install them together. The main motivation to avoid is for dependent packages, not for end-users. Unfortunately, the missing bit in Python, which would be convenient in this case, are _default extras_ (default features in Rust, as it is often `std`).

Yes I agree. In general I hate unnecessary dependencies, and it would be unfair to force those users to install something they didn't ask for. So having the CLI as an extra is a good compromise!

> 
> > on Windows this will lead to an error because there's a Python interface, but no CLI.
> 
> This would not be a problem. Dependencies can be specified as platform-dependent, and we could exclude Windows (until there will be a released wheel for that as well).

For the time being I suppose we can leave everything as is, if someone complains we know we have Windows users :smile:.

cschwan (approved):

Issue #270, opened 2024-04-05
[TODO] Add machine-readable MC results and uncertainties to calculate interpolation error


cschwan:
In [pinefarm](https://github.com/NNPDF/pinefarm) we add metadata to the generated PineAPPL grid. Unfortunately the metadata changed a bit over time and thus isn't uniform and also a bit hard to read automatically. An example of the `results` metadata for [`ATLAS_DY_7TEV_CF`](https://github.com/NNPDF/pineapplgrids/pull/19) is:

```
----------------------------------------------------------------------
   PineAPPL         MC        sigma      central         min      max 
                              1/100   sigma   1/1000   1/1000   1/1000
----------------------------------------------------------------------
 2.378766e+00  2.378884e+00   0.262   0.019   0.0496   0.1263   0.0397 
 6.711583e+00  6.711661e+00   0.086   0.014   0.0117   0.0855   0.0486 
 1.246710e+01  1.246338e+01   0.052   0.569   0.2981   0.2739   0.3157 
 1.930853e+01  1.930489e+01   0.012   1.525   0.1884   0.1806   0.2039 
 2.658518e+01  2.658270e+01   0.012   0.813   0.0935   0.0782   0.1117 
 3.351150e+01  3.351280e+01   0.007   0.534   0.0387   0.0357   0.0409 
 6.590712e+01  6.591164e+01   0.005   1.319   0.0685   0.0718   0.0676 
 3.044189e+01  3.044364e+01   0.011   0.534   0.0575   0.0428   0.0705 
 7.526862e+00  7.526933e+00   0.030   0.031   0.0094   0.0232   0.0019 
 1.890373e-01  1.890007e-01   0.079   0.245   0.1940   0.1845   0.2035
 4.149114e-01  4.148817e-01   0.019   0.378   0.0714   0.0634   0.0824
 6.575542e-01  6.575642e-01   0.010   0.149   0.0152   0.0105   0.0187
 6.353079e-01  6.353490e-01   0.010   0.631   0.0647   0.0660   0.0615
 2.540983e-01  2.541186e-01   0.018   0.433   0.0800   0.0767   0.0850
 4.145008e-02  4.145223e-02   0.061   0.085   0.0517   0.0390   0.0632
```

This can be improved:

1. Don't store the column `PineAPPL`; we can always calculate this number by convoluting the grid with the PDF given in the metadata `results_pdf`.
[...]

Issue #257, opened 2024-02-07
Investigate if we can compile/install the Fortran module


cschwan:
See this comment: <https://github.com/NNPDF/pineappl/pull/256#issuecomment-1931499353>.

cschwan:
After having written https://github.com/NNPDF/pineappl/pull/328#issuecomment-2633882463 I'm convinced that we probably shouldn't do this - then we have give stability guarantees which makes maintenance hard.

Issue #172, opened 2022-08-31
Generalize convolutions for both initial and final states


cschwan:
This generalization will require the following changes:

- [x] add support for new scale(s): fragmentation function scale. Requires changing the order structure, which is related to <https://github.com/NNPDF/pineappl/issues/98>. This is being addressed in #299.
- [x] the PDF-PDF-FF case will require a generalization of the `SparseArray3` structure; we need `SparseArray4`, and then it's possibly better to even think about `SparseArrayN`; this is done with #275.
- [x] what is currently called 'luminosities' should really haven been 'channels' from the start, and this part will need to be reworked to support the case of three PIDs, also for the case of three convolutions; partly addressed in commit aef098226ebf22e697106eb33177be2819d89b11, remainder requires file-format changes that are being addressed in #299.
- [x] the CLI needs to support two or three different PDFs: <https://github.com/NNPDF/pineappl/issues/174>
- [x] the evolution code (`Grid::evolve*`) needs to be changed in order to support time-like EKOs. For that to work we should discuss the possible interfaces and probably first merge <https://github.com/NNPDF/pineappl/pull/244>; partly addressed in #289. The remaining bits will be addressed in #299.

cschwan:
The implementation might take a while :smile:. But I anticipate to introduce a new file format soon to accommodate a number of changes/simplications (see #118) and it would be ideal to know what changes roughly are required to support them in said file format.

cschwan:
@enocera: which of the following scenarios do you see to be relevant in the future:

1. lepton-hadron collisions with one fragmentation function,
2. lepton-hadron collisions with two fragmentation functions,
3. hadron-hadron collisions with one fragmentation function,
4. hadron-hadron collisions with two fragmentation functions?

Basically my questions is: how many 'convolutions' (with PDFs and/or fragmentation functions) will we need to support in PineAPPL?

AleCandido:
I guess there is also SIA case, i.e. "lepton-lepton collisions with *".

cschwan:
What does SIA stand for?

enocera:
> I had a brief chat with @t7phy yesterday, and there are a couple of points that require answering:
> 
>     1. In which (file-)format are the fragmentation functions available? Is there a library like LHAPDF that supports it? Or can one (ab)use LHAPDF?

Sets of fragmentation functions are available through LHAPDF in the very same format as PDFs.
> 
[...]

PR #299, opened 2024-07-12
Implement v1 file format [WIP]

t7phy:
Feature TODOs:

- [x] Generalize `Channel` to support arbitrarily many PIDs. This is implemented with commits f32902afa8348525d3994ee968a0765c52675deb, 2417c9287b6ab12f3cde9ebd9787e5acb6778a69 and 03f2dc65c589a1668200923ffd20bfe543e32a86.
- [x] The structure `Order` should support a fragmentation scale. Started with commits 5e281aefc3fd5ea1015ae187b036937dc5bf17c0 and 02e48e9c7ce90a4e20d99e76043b1f731fa05c08.
- [x] The CLI should support varying this new fragmentation scale. We need the following point variations: `1` (no scale variation), `3` (vary all scales with the same factor), `7` and `9` (assuming that there's no fragmentation scale), and `17` and `27` (assuming there's a fragmentation scale)
- [x] Extend the evolution code to take care of (at least) one, two and three convolutions with possibly one, two and three different EKOs
- [x] Finally, we need new subgrid types that support possibly arbitrarily many scales and momentum fractions and which support filling or are optimized for space. We should use the `PackedArray` struct that we wrote some time ago.
- [x] To fix the tests, we need import `v0` PineAPPL grids and convert the subgrids into the new subgrid types.
- [x] Let the user specify on the CLI whether are convolution function is polarized or a fragmentation function; we can use a syntax like `LHAPDFNAME+p` for polarized and `LHAPDFNAME+f` for fragmentation function and `LHAPDFNAME+pf` or `LHAPDFNAME+fp` for both
- [ ] #270; started with commit 029f3a1

Code TODOs:

- [x] Change types of members of `Order` to `u8`
- [x] Generalize `channel!` to arbitrarily many PIDs
- [x] Support arbitrarily many scales and fix import of flexible-scale fastNLO tables
- [x] Test evolution for flexible-scale grids with `ZEUS_2JET_319GEV_374PB-1_DIF_ETQ2_BIN6` (available in the same location where also the other test data is)
- [x] Simplify bin treatment: remove `BinInfo`, and have instead fill limits (1d limits that only concern `Grid::fill`) and bin limits (n-dimensional limits)
- [x] Add fragmentation scales to `Grid::evolve_info`
- [x] Make it possible to have different initial factorization and fragmentation scales in an FK-table
- [x] Rename `lagrange_subgrid`
- [x] Add support for static scale detection in previous subgrid type
- [x] Rename `PackedQ1X2SubgridV1`
- [x] Remove `NodeValues`
- [x] Remove `Mu2`
- [x] Implement `D=N` for `general_tensor_mul` in `evolution.rs`
- [x] In `Grid::merge` check if grids are compatible with each other before merging
- [x] in the CLI remove query to `SetType:`
- [x] in the CLI don't assume polarization of the convolution functions from the grid
- [ ] Test `EvolveInfo::frg1` and `FkTable::frg0` @Radonirinaunimi #325 
- [ ] Test different initial factorization and fragmentation scales in an FK-table @Radonirinaunimi #325 
- [ ] Test `D=3` for `general_tensor_mul` in `evolution.rs` @Radonirinaunimi #325 
- [ ] Fill in all the TODO comments #326
- [ ] Cleanup code #326
- [ ] Raise code coverage #326

CAPI TODO:

- [x] increase code coverage
- [x] unbreak `GridOptFlags` (see commit 54a59f3f1216f055a60ed27f30e6bb0bf189d043)
- [x] rename `pineappl_channel_*` to `pineappl_channels_*`?
- [x] check generated grids of new example programs
- [x] rename `-v1.cpp` suffix of example programs to `.cpp` and rename the previous example programs from `.cpp` to `-deprecated.cpp`. This should clarify that the use of `v0` functions is deprecated (but still works!)

Fortran module TODO:

- [x] add new functions

Python API:

- [x] update to `pyo3-0.22.5`, this possible now since `numpy-0.22.0` was released. Figure out what new features we can leverage from `0.22.x` (see its [CHANGELOG](https://github.com/PyO3/pyo3/blob/main/CHANGELOG.md#0220---2024-06-24)):
  - [x] numpy-2 support (works out of the box)
  - [x] declarative modules and submodules
  - [x] Python 3.13
  - [x] recheck whether we still need `#![allow(unsafe_op_in_unsafe_fn)]`
- [x] raise minimum required Python version to `3.7` in `pyproject.toml`
- [x] propagate bin-related changes @Radonirinaunimi 
- [ ] Update tutorials #327

General:

- [x] merge `CHANGLOG.md` entries from `v0.8.x` branch
- [x] write a summary of the changes for v1.x


cschwan:
I believe the list shown in #118 is probably a bit too ambitious for this pull request. Realistically, we should implement all the items listed in #172.

pineappl/src/boc.rs
```diff
@@ -3,13 +3,63 @@
...
+    /// TODO
+    X(usize),
+}
+
+impl Kinematics {
+    /// TODO
+    pub const MU2_R: Self = Self::Mu2(Scale::REN);
+
+    /// TODO
+    pub const MU2_F: Self = Self::Mu2(Scale::FAC);
```
felixhekhorn, line 42: since you are already on it: you may want to consider an even more general case: there is µ for every convolution and not per convolution type. E.g. consider Unpol x Pol x Timelike - there are in principle three different mus, two of which one could call a factorization scale. However, there is no pQCD argument which forces them to be the same, although they are often _chosen_ to be the same. The same holds for unequal hadron collision, e.g. p-A at LHC: in principle I can choose an individual scale for each beam. For equal beams, e.g. p-p, this does not hold, as here things are symmetric by definition. (Of course the ren. scale is always there and separate)

cschwan: That's a good suggestion, and I had something even more general in mind which is that you separately set not only the scale of each convolution function (PDF, FF, ..), but also the scale of different couplings (alphas, scale-dependent alpha, [other couplings in MS-bar renormalized EFTs](https://github.com/NNPDF/pineappl/issues/98), ...). But all of this requires a bit more thinking which will have to be done afterwards, with lower priority. The highest priority for the time being is to get this branch merged and to make 3 convolutions working.

pineappl_py/src/bin.rs
```diff
@@ -16,13 +16,19 @@ pub struct PyBinRemapper {
...
     /// Parameters
     /// ----------
     /// normalizations : list(float)
     ///     bin normalizations
     /// limits : list(tuple(float, float))
     ///     bin limits
     #[new]
+    #[must_use]
+    #[allow(clippy::needless_pass_by_value)]
     pub fn new(normalizations: PyReadonlyArray1<f64>, limits: Vec<(f64, f64)>) -> Self {
```
cschwan, line 32: Change `PyReadonlyArray1` to `Vec` and remove `#[allow(..)]`.

pineappl_py/src/boc.rs
```diff
@@ -126,6 +148,8 @@ impl PyOrder {
     /// numpy.ndarray(bool)
     ///     boolean array, to be used as orders' mask
     #[staticmethod]
+    #[must_use]
+    #[allow(clippy::needless_pass_by_value)]
```
cschwan, line 152: We shouldn't suppress the argument, but instead fix the source of the problem.

Radonirinaunimi:
For some reasons I did not receive notifications from github that some changes were/are happening, so I got a bit of conflicts when pulling last night (due the Python files being modified at the same time).

@cschwan These points have been addressed in some parts now (I believe your comments also happened before we had a chat? xd). I am still polishing a bit the interface and fixing/improving the tests and at the same time fixing the docs as well. I will push everything once done (hopefully very soon).

pineappl_py/src/import_subgrid.rs
```diff
@@ -23,63 +24,28 @@ impl PyImportSubgridV1 {
...
-        array: PyObject,
-        scales: Vec<f64>,
-        x1_grid: Vec<f64>,
-        x2_grid: Option<Vec<f64>>,
-        py: Python<'py>,
-    ) -> Self {
-        // let node_values: Vec<Vec<f64>> = vec![scales, x1_grid, x2_grid];
-        // let mut sparse_array: PackedArray<f64> =
-        //     PackedArray::new(node_values.iter().map(Vec::len).collect());
+    pub fn new(array: PyReadonlyArrayDyn<f64>, scales: Vec<f64>, x_grids: Vec<Vec<f64>>) -> Self {
```
cschwan, line 35: @Radonirinaunimi merge `scales` and `x_grids` in a single `node_values: Vec<Vec<f64>>`, because we may have more than two scales (flexible-scale fastNLO tables are an example).

pineappl_capi/src/lib.rs
```diff
@@ -1357,3 +1355,253 @@ pub unsafe extern "C" fn pineappl_string_delete(string: *mut c_char) {
...
+
+/// Adds a generalized linear combination of initial states to the Luminosity.
+///
+/// # Safety
+///
+/// The parameter `lumi` must point to a valid `Lumi` object created by `pineappl_lumi_new`.
+/// `pdg_id_combinations` must be an array with length `n_combinations * combinations`, and
+/// `factors` with length of `combinations`.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_lumi_add2(
```
cschwan, line 1426: Instead of `lumi`, let's use `channel` and then you can drop the `2`.

pineappl_capi/src/lib.rs
```diff
@@ -1357,3 +1355,253 @@ pub unsafe extern "C" fn pineappl_string_delete(string: *mut c_char) {
...
+    for k in 0..=nb_convolutions {
+        let kin = unsafe {
+            let kin_ptr = *kinematics.add(k);
+            CStr::from_ptr(kin_ptr)
+                .to_str()
+                .expect("Failed to convert one of the `kinematics` types into a proper string.")
+        };
+
+        let kin_type = match kin {
+            "scale" => Kinematics::Scale(0),
```
cschwan: This we must make more general, because we'll also need at least `Kinematics::Scale(1)` for flexible-scale fastNLO tables.

pineappl_capi/src/lib.rs
```diff
@@ -1357,3 +1355,253 @@ pub unsafe extern "C" fn pineappl_string_delete(string: *mut c_char) {
...
+
+    Box::new(Grid::new(
+        pid_basis,
+        channels.0.clone(),
+        orders,
+        bin_limits,
+        convolutions,
+        interp_vecs,
+        kinematics_vec,
+        Scales {
```
cschwan, line 1573: This the user should also be able to set, because we might have completely different values for renormalization, factorization and fragmentation scales.

pineappl_capi/src/lib.rs
```diff
@@ -1357,3 +1355,253 @@ pub unsafe extern "C" fn pineappl_string_delete(string: *mut c_char) {
...
+/// this function is not safe to call.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_grid_fill2(
+    grid: *mut Grid,
+    order: usize,
+    observable: f64,
+    lumi: usize,
+    ntuple: *const f64,
+    weight: f64,
+    size_tuple: usize,
```
cschwan, line 1596: Omit `size_tuple` and instead we must assume that this is given by `grid.kinematics().len()`.

pineappl_capi/src/lib.rs
```diff
@@ -1357,3 +1355,253 @@ pub unsafe extern "C" fn pineappl_string_delete(string: *mut c_char) {
...
+/// # Safety
+///
+/// If `grid` does not point to a valid `Grid` object, for example when `grid` is the null pointer,
+/// this function is not safe to call.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_grid_fill2(
+    grid: *mut Grid,
+    order: usize,
+    observable: f64,
+    lumi: usize,
```
cschwan, line 1593: Rename `lumi` to `channel`.

pineappl_capi/src/lib.rs
```diff
@@ -1357,3 +1355,253 @@ pub unsafe extern "C" fn pineappl_string_delete(string: *mut c_char) {
...
+// Here starts the generalized C-API interface.
+
+/// Type for defining the interpolation object
+#[repr(C)]
+pub struct InterpTuples {
+    node_min: f64,
+    node_max: f64,
+    nb_nodes: usize,
+    interp_degree: usize,
+    reweighting_method: *const c_char,
```
cschwan, line 1368: For the three members that are `*const c_char` it would nice if we could use directly or indirectly use the `enum`s from Rust and use them in C. Maybe a
```rust
pub type InterpMeth = pineappl::interpolation::InterpMeth;
```
works? If `#[repr(C)]` is needed we can add it in the Rust crate.

pineappl_capi/src/lib.rs
Radonirinaunimi: Thanks for the comments @cschwan! I will address them.

As for this particular one, it is already superseded by this commit: [5ff86af](https://github.com/NNPDF/pineappl/pull/299/commits/5ff86af1d6c649336ed4cbf44c1badb591dfe6ee) which indeed makes it more general.

cschwan: For `x` yes, but I don't see how `Scale(1)` can be constructed.

pineappl_capi/src/lib.rs
```diff
@@ -1357,3 +1355,253 @@ pub unsafe extern "C" fn pineappl_string_delete(string: *mut c_char) {
...
+/// # Safety
+///
+/// # Panics
+#[no_mangle]
+#[must_use]
+pub unsafe extern "C" fn pineappl_grid_new2(
+    pid_basis: *const c_char, // takes: `const char* pid_basis = "Evol";`
+    channels: *const Lumi,
+    orders: usize,
+    order_params: *const u32,
```
cschwan, line 1484: Let's use `u8` instead of `u32` for `order_params`, then you can get rid of the `.try_into().unwrap()` below.

pineappl_capi/src/lib.rs
Radonirinaunimi: Addressed in [4d0f526](https://github.com/NNPDF/pineappl/pull/299/commits/4d0f526634fa0c2e40966ba30612d15dcf4fed15). The other comments I marked resolved instead have been addressed in [93d4d62](https://github.com/NNPDF/pineappl/pull/299/commits/93d4d625decc4206b7073397797a4c3ac2f20253) but please check (I mainly resolved them in order to know myself which one I've addressed xp ). I also added another "temporary" example to check this.

The remaining comments I still have to address.

pineappl_capi/src/lib.rs
Radonirinaunimi: This is working perfect for `InterpTuples` - and the code is much cleaner. Now, trying to do the same for the others.

Radonirinaunimi: Addressed in [016eb4b](https://github.com/NNPDF/pineappl/pull/299/commits/016eb4bba61fb40d1fa3c0cc272d59667c499ffe). I wanted to promote this:
https://github.com/NNPDF/pineappl/blob/016eb4bba61fb40d1fa3c0cc272d59667c499ffe/pineappl_capi/src/lib.rs#L1462

to `ScaleFuncForm` but the enumerator `Scale` clashes with `Kinematic::Scale`.

PS: Sorry I re-introduced the wrong output when doing `git pull --rebase`.

pineappl_py/src/import_subgrid.rs
Radonirinaunimi: Addressed in [2a36294](https://github.com/NNPDF/pineappl/pull/299/commits/2a362946a3d7aee77475fa69b2596bb65fb9cf47).

Radonirinaunimi:
So to summarize a little bit the status so far (as things might be amalgamated within the commits):

- AFAIK, the **C** API is almost done, what is really missing is only the generic `convolution`. Rust's `enum`s are now used whenever possible - simplifying quite a bit the implementations (https://github.com/NNPDF/pineappl/pull/299/commits/016eb4bba61fb40d1fa3c0cc272d59667c499ffe). The corresponding examples/tests, however, still need to be polished/implemented. But the main should already be there.
- The **Fortran** API is little bit behind. Currently, it is still missing the implementation of the generic `convolution` and `pineappl_grid_new`. The latter is a little bit involved (although not complicated) as it requires translating all the C `enum`s. As in all the cases, examples and tests also need to be implemented.
- As for the **Python** API, everything should be implemented **correctly** except for `grid.evolve`. 
> update to pyo3-0.22.5, this possible now since numpy-0.22.0 was released. Figure out what new features we can leverage from 0.22.x (see its [CHANGELOG](https://github.com/PyO3/pyo3/blob/main/CHANGELOG.md#0220---2024-06-24)):

A major feature that we can use now (which we couldn't do before) is having `#[pyclass]` on `enum`s that have (complex) tuple variants. This means that we directly interface the Rust `enum`'s objects, hence making the interface more robust. This has been implemented for example in [2a36294](https://github.com/NNPDF/pineappl/pull/299/commits/2a362946a3d7aee77475fa69b2596bb65fb9cf47) and [48d6f26](https://github.com/NNPDF/pineappl/pull/299/commits/48d6f267548a5f818bf983dd6c3bcf6b0f77ef67).

> recheck whether we still need #![allow(unsafe_op_in_unsafe_fn)]

This I don't think we'll ever be able to get rid of. Most of the C parts that Python rely on are inherently unsafe from Rust perspectives. 

> declarative modules and submodules

What exactly do you mean by this @cschwan (in comparison to what's been done so far)?

All in all, I hope to finish these remaining todos in the next days (unfortunately I might not be able to look into it Today (Monday)).

felixhekhorn:
> > declarative modules and submodules
> 
> What exactly do you mean by this @cschwan (in comparison to what's been done so far)?

I think @cschwan is thinking of [this syntax](https://pyo3.rs/v0.22.5/module#declarative-modules) instead of the [register syntax](https://pyo3.rs/v0.22.5/module#python-submodules) that we introduced in #302

pineappl_py/tests/test_grid.py
```diff
@@ -341,6 +341,49 @@ def test_fill(self):
...
+
+        # Convolution of two symmetrical hadrons
+        h = ConvType(polarized=False, time_like=False)
+        h_conv = Conv(conv_type=h, pid=2212)
+        res = g.convolve(
+            pdg_convs=[h_conv, h_conv],
+            xfxs=[lambda pid, x, q2: x, lambda pid, x, q2: x],
+            alphas=lambda q2: 1.0,
+        )
+        pytest.approx(res) == 0.0
```
felixhekhorn, line 366: we should not test against `0.0`, but against something else (as we want to see something is actually inside the grid)

cschwan:
>     * The **Fortran** API is little bit behind. Currently, it is still missing the implementation of the generic `convolution` and `pineappl_grid_new`. The latter is a little bit involved (although not complicated) as it requires translating all the C `enum`s. As in all the cases, examples and tests also need to be implemented.

I was thinking maybe @janw20 would be interested in taking care of this.

cschwan:
> > declarative modules and submodules
> 
> What exactly do you mean by this @cschwan (in comparison to what's been done so far)?

There are two roughly things:
1. the submodule `register` functions add some manual `import sys; sys.modules['...'] = m`. I wonder whether this has been fixed/added to the new PyO3 version
2. If that's the case, we could try to completely get rid of the `register` function by using the declarative syntax.

pineappl/src/boc.rs
```diff
@@ -571,12 +571,12 @@ impl Channel {
...
     /// use pineappl::channel;
     ///
-    /// let entry = channel![103, 11, 10.0].translate(&|evol_id| match evol_id {
+    /// let entry = channel![10.0 * (103, 11)].translate(&|evol_id| match evol_id {
     ///     103 => vec![(2, 1.0), (-2, -1.0), (1, -1.0), (-1, 1.0)],
     ///     _ => vec![(evol_id, 1.0)],
     /// });
     ///
-    /// assert_eq!(entry, channel![2, 11, 10.0; -2, 11, -10.0; 1, 11, -10.0; -1, 11, 10.0]);
+    /// assert_eq!(entry, channel![10.0 * (2, 11) + -10.0 * (-2, 11) + -10.0 * (1, 11) + 10.0 * (-1, 11)]);
```
felixhekhorn, line 579: I think you need to stress more that a literal `+` is needed, i.e. `10.0 * (2,11) - 10.0 * (-2, 11)` will not work (or support that as well), as people are used to "normal" math (where we precisely define that shortcut :upside_down_face: )

cschwan: I'm thinking of actually removing the macro altogether because it's only used in Rust, and there even only as a shortcut.

felixhekhorn: well macros are in the first place short cuts, no?

janw20:
> @janw20, are you interested in taking care of this?

Yes, I can have a look :+1: If I understand correctly from reading the code, the plan for the C interface is that all function names are backward compatible? (So for new functionality, the names are either appended with a 2 or get a completely new name?)

cschwan:
> > @janw20, are you interested in taking care of this?
> 
> [..] If I understand correctly from reading the code, the plan for the C interface is that all function names are backward compatible? (So for new functionality, the names are either appended with a 2 or get a completely new name?)

Yes, everything you've written so far shouldn't should still compile with the CAPI. For the Fortran interface use whatever names we have in the CAPI.

pineappl_py/tests/test_grid.py
Radonirinaunimi: Yes, absolutely! These were parts of the legacy checks that were there before. But we'll definitely improve these.

Radonirinaunimi:
> There are two roughly things:
> 
>     1. the submodule `register` functions add some manual `import sys; sys.modules['...'] = m`. I wonder whether this has been fixed/added to the new PyO3 version
> 
>     2. If that's the case, we could try to completely get rid of the `register` function by using the declarative syntax.

Ok, got it! I will check this.

Radonirinaunimi:
Thanks @janw20 for the helps! Everything that requires Fortran implementation is below this line:
https://github.com/NNPDF/pineappl/blob/934d153e276bafa0659d6ca86f733a5a2aae09e4/pineappl_capi/src/lib.rs#L1359

Some of them I already implemented but it is still missing the crucial main components, namely `pineappl_grid_new2` and `pineappl_convolve`.

pineappl_py/src/grid.rs
```diff
@@ -326,8 +342,9 @@ impl PyGrid {
     ///
     /// Parameters
     /// ----------
-    /// slices : Iterable
-    ///     list of (PyOperatorSliceInfo, 5D array) describing each convolution
+    /// slices : list(list(tuple(PyOperatorSliceInfo, PyReadOnlyArray4)))
+    ///     list of EKOs where each element is a list of (PyOperatorSliceInfo, 5D array)
```
felixhekhorn, line 346:
```suggestion
    ///     list of EKOs where each element is a list of (PyOperatorSliceInfo, 4D array)
```
?

Radonirinaunimi: Yes, it is indeed a 4D,

Btw, this is what I briefly asked in yesterday's CM re `pineko` (to which I now know the answer). For the time being, each operator is passed as a vector but it'll be changed to a generator.

pineappl/src/interpolation.rs
```diff
@@ -713,4 +716,42 @@ mod tests {
...
+            40,
+            3,
+            ReweightMeth::NoReweight,
+            Map::ApplGridH0,
+            InterpMeth::Lagrange,
+        )
+        .node_values();
+
+        for &q2 in &q2_ref {
+            // these two functions should are inverse to each other, within numerical noise
```
felixhekhorn, line 753:
```suggestion
            // these two functions should be inverse to each other, within numerical noise
```

pineappl_py/tests/test_grid.py
Radonirinaunimi: Done in [84a4e97](https://github.com/NNPDF/pineappl/pull/299/commits/84a4e9771ba5563e7f9b8edf9e12f3ab1bfb3b18).

pineappl_py/tests/test_boc.py
```diff
@@ -10,11 +11,53 @@ def test_init(self):
 
 
 class TestKinematics:
-    def test_init(self):
-        kin = Kinematics.Scale(0)
-        scale_func = ScaleFuncForm.Scale(0)
-        assert isinstance(kin, Kinematics)
-        assert isinstance(scale_func, ScaleFuncForm)
+    @pytest.mark.parametrize(
+        "kintype, argument, expected_type",
```
felixhekhorn, line 15: the third element seems to be always the same - I guess we can hardcode it, no? same for `ScaleFuncForm`

Radonirinaunimi: Done in [e7c2e88](https://github.com/NNPDF/pineappl/pull/299/commits/e7c2e880c776ed620668ef8b2954cc9853961db2).

pineappl_py/docs/source/introduction.ipynb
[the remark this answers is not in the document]
Radonirinaunimi: You're right, `la` is more convenient for consistency. Done in [e7c2e88](https://github.com/NNPDF/pineappl/pull/299/commits/e7c2e880c776ed620668ef8b2954cc9853961db2). I also addopted $\ell_A$ when referring to the log (I am also not sure that is the default name for fragmentation scale).

janw20:
Hi, I completed the Fortran interface. I made some (minor) design decisions that I would like your opinion on:

- Since `Kinematics` is an enum with tuple-like variants, it is converted to a C struct holding a tag variable and a C union. Fortran doesn't support unions, so I use a Fortran derived type (`pineappl_kinematics`) that also has a tag variable, but only one integer variable for the index instead of the C union. I found this to be the cleanest solution, or do you have a better idea?
- For some reason, Fortran cannot do arrays of procedure pointers, so I created a new derived type for the PDF and alpha_s functions, `pineappl_xfx` and `pineappl_alphas`, that each holds the procedure pointer as a member named `proc`. This is needed so that an array of PDF functions can be passed to `pineappl_grid_convolve`. For consistency, I also changed this in `pineappl_grid_convolve_with_one` and `pineappl_grid_convolve_with_two`. This goes against keeping backward compatibility with the 0.x of Pineappl versions, but until there is a solution to #257, I figured that the user would make their own copy of `pineappl.f90` anyway and use that, so backward compatibility is not that important in the Fortran case.

I also noticed two small bugs in the v1 C++ examples which I fixed in my first commit. Related to that: Is there a reason why `pineappl_grid_new` creates the kinematics in the order (Q², x₁, x₂), and not (x₁, x₂, Q²), as was always the order in, e.g., `pineappl_grid_fill`? I think this might have been the reason for the bug in [examples/cpp/fill-grid-v1.cpp](../blob/master/examples/cpp/fill-grid-v1.cpp) line 113

Radonirinaunimi:
Hi @janw20, thanks a lot for completing the Fortran interface!

> Hi, I completed the Fortran interface. I made some (minor) design decisions that I would like your opinion on:
> 
>     * Since `Kinematics` is an enum with tuple-like variants, it is converted to a C struct holding a tag variable and a C union. Fortran doesn't support unions, so I use a Fortran derived type (`pineappl_kinematics`) that also has a tag variable, but only one integer variable for the index instead of the C union. I found this to be the cleanest solution, or do you have a better idea?
> 
>     * For some reason, Fortran cannot do arrays of procedure pointers, so I created a new derived type for the PDF and alpha_s functions, `pineappl_xfx` and `pineappl_alphas`, that each holds the procedure pointer as a member named `proc`. This is needed so that an array of PDF functions can be passed to `pineappl_grid_convolve`. For consistency, I also changed this in `pineappl_grid_convolve_with_one` and `pineappl_grid_convolve_with_two`. This goes against keeping backward compatibility with the 0.x of Pineappl versions, but until there is a solution to [Investigate if we can compile/install the Fortran module #257](https://github.com/NNPDF/pineappl/issues/257), I figured that the user would make their own copy of `pineappl.f90` anyway and use that, so backward compatibility is not that important in the Fortran case.
> 
I had a look at it and it looks good. I don't think there is a cleanest a more robust to approach this apart from what is being done now (actually, this was how I also envisioned how it'd look like).
> 
> I also noticed two small bugs in the v1 C++ examples which I fixed in my first commit. 
>

Thanks for spotting and fixing this bug! I had this under my radar before but somehow forgot to fix it (maybe I remembered to only do so on the Python API).

> Related to that: Is there a reason why `pineappl_grid_new` creates the kinematics in the order (Q², x₁, x₂), and not (x₁, x₂, Q²), as was always the order in, e.g., `pineappl_grid_fill`? I think this might have been the reason for the bug in [examples/cpp/fill-grid-v1.cpp](../blob/master/examples/cpp/fill-grid-v1.cpp) line 113
>
I don't think there is a real reason why the kinematics have to be ordered in such a way. The restriction is coming from these lines:
https://github.com/NNPDF/pineappl/blob/63dc010000a27ccb94e02923127be7c765de9a7c/pineappl/src/grid.rs#L281-L285
which has a `TODO`, so likely this will change.

pineappl_py/tests/test_grid.py
```diff
@@ -212,6 +212,55 @@ def test_bins(self, fake_grids):
...
+    def test_rotate_pidbasis(self, fake_grids):
+        g = fake_grids.grid_with_generic_convolution(
+            nb_convolutions=2,
+            channels=CHANNELS,
+            orders=ORDERS,
+            convolutions=[CONVOBJECT, CONVOBJECT],
+        )
+        # Rotate the Grid into the PDG basis
+        g.rotate_pid_basis(PidBasis.Pdg)
+        assert g.pid_basis == PidBasis.Pdg
```
felixhekhorn, line 225: you need to check the convolution numbers stay the same

pineappl_cli/tests/convolve.rs
```diff
@@ -15,8 +15,6 @@ Options:
   -b, --bins <BINS>       Selects a subset of bins
   -i, --integrated        Show integrated numbers (without bin widths) instead of differential ones
   -o, --orders <ORDERS>   Select orders manually
-      --xir <XIR>         Set the variation of the renormalization scale [default: 1.0]
-      --xif <XIF>         Set the variation of the factorization scale [default: 1.0]
```
felixhekhorn, line 17: is this a temporary change? because this feature is actually quite useful

janw20:
Just for the record, the panicking of the Fortran v1 examples could be resolved by explicitly declaring the constants given to `pineappl_interp_tuples` as double precision (i.e., suffixing them with `_dp`):
https://github.com/NNPDF/pineappl/blob/f478e29a233d2276e7eb726d6788a7f0e221a389/examples/fortran/test_v1.f90#L50-L52

Without this, [`applgrid::fx2`](https://github.com/NNPDF/pineappl/blob/dddf8250e1402bc004ef90d6e3f9f8b4447f699f/pineappl/src/interpolation.rs#L18) didn't converge for higher `y` values, which resulted in a panic when reaching [`unreachable!()`](https://github.com/NNPDF/pineappl/blob/dddf8250e1402bc004ef90d6e3f9f8b4447f699f/pineappl/src/interpolation.rs#L33).

pineappl_cli/tests/convolve.rs
cschwan: Maybe I'm mistaken, but it seems @Radonirinaunimi removed the feature during the merge. You should also provide the possibility to vary to fragmentation scale.

Radonirinaunimi:
Thanks again @janw20 for fixing this issue! I still don't understand why this was only (?) happening on Arch, but I guess this doesn't matter anymore.

pineappl_py/tests/test_grid.py
Radonirinaunimi: Addressed in [adeb134](https://github.com/NNPDF/pineappl/pull/299/commits/adeb134960eb112e6d72e32e1ae6c1721a62f435).

Radonirinaunimi:
> @Radonirinaunimi and @janw20: could one of you take care of the new item "check generated grids of new example programs"?

Sorry @cschwan for the very late reply, I was away for a while. Yes, I will check while I am finishing the CAPI TODO.

Radonirinaunimi:
For the record, the Python installation failed with maturin `v1.8.0` because of this PR https://github.com/PyO3/maturin/pull/2391. But simply adding `dynamic = ["version"]` to the `pyproject.toml` solved the issue (this shouldn't have any effect when releasing to PyPI).

cschwan:
Great :+1:, thank you @Radonirinaunimi! Are all the grids that are being generated by the examples also being convolved? If that's the case you can also tick the corresponding checkbox.

cschwan:
> For the record, the Python installation failed with maturin `v1.8.0` because of this PR [PyO3/maturin#2391](https://github.com/PyO3/maturin/pull/2391). But simply adding `dynamic = ["version"]` to `pyproject.toml` solved the issue (this shouldn't have any effect when packaging to PyPI).

Since no one of us knew this, we could document that this is a requirement. Our requirement is that we want version numbers only in `Cargo.toml` files and therefore all other version numbers must be dynamically derived; I found this link: <https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#static-vs-dynamic-metadata>.

Radonirinaunimi:
Happy new year all!

> Great 👍, thank you @Radonirinaunimi! Are all the grids that are being generated by the examples also being convolved? If that's the case you can also tick the corresponding checkbox.

This is now the case. There is only one item left to fully finish the C API part.

> Since no one of us knew this, we could document that this is a requirement. Our requirement is that we want version numbers only in Cargo.toml files and therefore all other version numbers must be dynamically derived; I found this link: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#static-vs-dynamic-metadata.

Yes, we definitely should document this.

.github/workflows/python.yml
```diff
@@ -7,7 +7,8 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    # `ubuntu-22.04` doesn't support Python 3.7 anymore
```
felixhekhorn, line 10: The commit message and this comment seem to contradict each other ...

cschwan: Fixed in commit 6bcfc1cce4e4bb03e4e48ea818f357b3b36d9593.

CHANGELOG.md
```diff
@@ -38,6 +38,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - raised MSRV to 1.80.1
 - changed the order of elements in `Grid::fill` of the parameter `ntuple` to
   reflect the ordering of `kinematics` given to `Grid::new`
+- renamed the following switches of `pineappl write`: `--remap` to
+  `--set-bins`, `--remap-norm-ignore` to `--div-bin-norm-dims` and
+  `--remap-norm` to `--mul-bin-norm`. These names should be reflect the
```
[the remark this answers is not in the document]
cschwan: Cheers!

pineappl/src/boc.rs
```diff
@@ -209,6 +212,460 @@ impl Scales {
...
+        self.bins
+            .first()
+            // UNWRAP: `Bin::new` should guarantee that there's at least one bin
+            .unwrap_or_else(|| unreachable!())
+            .dimensions()
+    }
+
+    /// TODO
+    #[must_use]
+    pub fn fill_index(&self, value: f64) -> Option<usize> {
```
felixhekhorn, line 381: this prefix is a bit confusing, i.e. `fill_index` and `fill_limits` are not similar to each other - maybe `insert_at_index`?

felixhekhorn: or actually, just `insert`?

pineappl/src/boc.rs
```diff
@@ -209,6 +212,460 @@ impl Scales {
...
+
+        Ok(Self::from_limits_and_normalizations(limits, normalizations)
+            // TODO: implement proper error handling
+            .unwrap())
+    }
+}
+
+/// Error type returned by [`BinRemapper::from_str`]
+#[derive(Debug, Error)]
+pub enum Bla {
```
felixhekhorn, line 663:
```suggestion
pub enum BinRemapperStrError {
```
nur um das klar zu stellen: mein Standardname is 50% `Bla` und 50% `Blub` :see_no_evil:

pineappl/src/grid.rs
```diff
@@ -124,17 +124,17 @@ bitflags! {
 #[derive(Clone, Deserialize, Serialize)]
 pub struct Grid {
     subgrids: Array3<SubgridEnum>,
-    channels: Vec<Channel>,
-    bin_limits: BinLimits,
+    bwfl: BinsWithFillLimits,
```
felixhekhorn, line 127: Can't we call it just `bins` here please? We don't care about the implementation details inside grid

pineappl/src/boc.rs
```diff
@@ -209,6 +212,460 @@ impl Scales {
...
+            .iter()
+            .zip(other.bins())
+            .all(|(lhs, rhs)| lhs.partial_eq_with_ulps(rhs, ulps))
+    }
+}
+
+impl FromStr for BinsWithFillLimits {
+    type Err = Bla;
+
+    fn from_str(s: &str) -> Result<Self, Bla> {
```
felixhekhorn, line 512: Is it possible to split this looong function into 4-5 smaller ones? naively it seems there is a clear path step 1 -> step 2 -> step 3 -> ... which act in isolation

pineappl/src/boc.rs
cschwan: They are similar to each other in that they both refer to `Grid::fill`. This method has a parameter `observable` which is an `f64`, which is being checked if it falls inside the limits returned by `BinsWithFillLimits::fill_limits`. If that's not the case, `BinsWithFillLimits::fill_index` returns `None`, else `Some(index)` where `index` is the result. Of course, this should be part of the documentation that is still missing.

pineappl/src/boc.rs
cschwan: Yes, that still has to be changed.

pineappl/src/grid.rs
cschwan: It's not an implementation detail. Every bin has effectively two sets of limits:
1. The N-dimensional actual bin limits that also the user sees. These mustn't overlap, but they allow 'gaps' between them, also they must all have the same dimensions;
2. Fill limits: these are one-dimensional limits, the 'fill limits', that only `Grid::fill` uses. These are as many limits as there are bins plus one more and all values must be ascending.

cschwan: That being said I suppose we can find a better name. I'd like to avoid `Bins` though, because this struct has a method `fn bins()` and then the code would have a lot of `grid.bins().bins()` invocations, which is a bit odd.

pineappl/src/boc.rs
cschwan: I can split this function into smaller ones, but I don't see that making it more readable. But let me try to improve it.

pineappl/src/grid.rs
felixhekhorn: > That being said I suppose we can find a better name.

that was my main concern - `bins_config`?

pineappl/src/boc.rs
felixhekhorn: to me it would be more readable - because then there will be a clear input, a clear task, and a clear output. Needless to say that this would also be good for unit testing.

pineappl/src/boc.rs
felixhekhorn: okay, I misunderstood (so yes documentation) ... mmm I still think it's not a perfect name (as usual this is a subjective thing); how about `search_fill_index`? if you disagree please just resolve this comment

pineappl/src/boc.rs
Radonirinaunimi: I agree that if this could be broken into smaller functions that'd be great as right now it is a bit difficult to read.

PR #301, opened 2024-07-24
Update Python docs

felixhekhorn:
Closes #157 

This mostly updates and fixes all Python doc strings. 

Apart from that there are few more (even breaking) changes:
- `PyGrid::evolve` got deprecated (as the underlying function)
- `PyFkTable::convolve_with_two`: return type was harmonized with `PyFkTable::convolve_with_one` to avoid the deprecation of `into_pyarray`
- the transition "Lumi" -> "Channel" was performed both on the PyO3 and Python side

if necessary we could also add other changes to the Python interface here, such as [to expose `Grid::convolutions`](https://github.com/NNPDF/pineko/pull/181#discussion_r1689352151)

Issue #200, opened 2022-12-16
Improve Python interface


alecandido:
There are a bunch of items that can be improved in the Python interface, eventually leading to less code to maintain and better DX for `pineappl_py` consumers:

- [x] drop Python layer, i.e. `pineappl_py/pineappl`, since at this point we have enough tools in PyO3 ecosystem ([`pyo3`](https://github.com/PyO3/pyo3), [`rust-numpy`](https://github.com/PyO3/rust-numpy), and extras, e.g. [logging](https://pyo3.rs/v0.17.3/ecosystem/logging.html), if ever any of these will be needed) to provide a Python native experience without a dedicated layer (we can work with numpy arrays, add static/classmethods, and getters and setters with properties)
- [x] code generation: a lot of bindings are really repetitive, and it is clear we can automate many of them; PyO3 itself is [based on code generation](https://pyo3.rs/v0.17.3/rust_cpython.html#macros), but mainly targeting Rust-powered Python libs, not bindings for standalone Rust libs, so we should clutter `pineappl` crate with PyO3 declarative macros; this is rather ugly, because then we might want to do the same for C bindings, and further ones in case we will ever support new languages (so scaling the number of macros with the languages supported); so either we find a native way to generate code using PyO3 macros, or we should write our own macros to generate the code corresponding to repeated patterns in current content of `pineappl_py/src`
- [ ] Python types: types are useful in Rust as in Python, so, if possible, let's propagate them to the second, even though PyO3 still indicates [a manual approach](https://pyo3.rs/v0.17.3/python_typing_hints.html)
- [ ] better Python exceptions: we can use PyO3 support for [plain exceptions](https://pyo3.rs/v0.17.3/exception.html), or its [`anyhow` feature](https://pyo3.rs/v0.17.3/features.html#anyhow), since @cschwan already started introducing `anyhow` in main `pineappl` crate

cschwan:
Here's another take on this: <https://towardsdatascience.com/nine-rules-for-writing-python-extensions-in-rust-d35ea3a4ec29>.

AleCandido:
> Here's another take on this: [towardsdatascience.com/nine-rules-for-writing-python-extensions-in-rust-d35ea3a4ec29](https://towardsdatascience.com/nine-rules-for-writing-python-extensions-in-rust-d35ea3a4ec29).

I need some time to digest. If you already did (or you will do before), can you turn the content of the article in further points in the OP? (or a separate message if you like more)

[...]

PR #302, opened 2024-08-06
Use PyO3 functionality to remove Python files

Radonirinaunimi:
Towards addressing https://github.com/NNPDF/pineappl/issues/200.

TODO:
- [x] improve and/or add documentations $\leftarrow$ done through #301 
- [x] rename instances of `*lumi*` into `*channel*` to be consistent with the `pineappl` crate $\leftarrow$ done through #301


cschwan:
Thank you for taking care of this, @Radonirinaunimi! Whenever you need help, let me know!

felixhekhorn:
> Thank you for taking care of this, @Radonirinaunimi! Whenever you need help, let me know!

same here!

Radonirinaunimi:
Although fixing and/or adding some documentations and doing more tests are still to be done, this is mostly ready (in that it passes all the tests) so I think you can already have a look.

There are only a few things that are not exactly the same as before (and that we may or may not want to address):
- the methods `.create()` and `.from_grid()` are no longer available, instead objects are instantiated from the class
- `.into()` needs to be manually called on subgrids in order to convert into `SubgridEnum`

PS: I could not force-push so please ignore the commit history 😇

pineappl_py/src/bin.rs
```diff
@@ -6,7 +6,7 @@ use pyo3::prelude::*;
 /// PyO3 wrapper to :rustdoc:`pineappl::bin::BinRemapper <bin/struct.BinRemapper.html>`
 ///
 /// **Usage**: `yadism`
-#[pyclass]
+#[pyclass(name = "BinRemapper")]
```
cschwan, line 9: Can't we name the class `BinRemapper` directly?

pineappl_py/src/grid.rs
```diff
@@ -383,15 +383,16 @@ impl PyGrid {
     ///     numpy.ndarray(float) :
     ///         cross sections for all bins, for each scale-variation tuple (first all bins, then
     ///         the scale variation)
+    #[pyo3(signature = (pdg_id, xfx, alphas, order_mask = None, bin_indices = None, lumi_mask= None, xi = None))]
```
cschwan, line 386: Add one space between `lumi_mask` and `=`.

pineappl_py/src/grid.rs
```diff
@@ -383,15 +383,16 @@ impl PyGrid {
...
     pub fn convolve_with_one<'py>(
         &self,
         pdg_id: i32,
         xfx: &Bound<'py, PyAny>,
         alphas: &Bound<'py, PyAny>,
-        order_mask: PyReadonlyArray1<bool>,
-        bin_indices: PyReadonlyArray1<usize>,
-        lumi_mask: PyReadonlyArray1<bool>,
-        xi: Vec<(f64, f64)>,
+        order_mask: Option<PyReadonlyArray1<bool>>,
```
cschwan, line 392: I noticed an inconsistency (?) that was introduced a long time ago, we use both `PyReadonlyArray1` and `Vec`. Does anyone happen to know the difference on the Python side?

cschwan (changes requested):

pineappl_py/src/grid.rs
```diff
@@ -447,17 +448,18 @@ impl PyGrid {
     ///     numpy.ndarray(float) :
     ///         cross sections for all bins, for each scale-variation tuple (first all bins, then
     ///         the scale variation)
+    #[pyo3(signature = (pdg_id1, xfx1, pdg_id2, xfx2, alphas, order_mask = None, bin_indices = None, lumi_mask= None, xi = None))]
```
cschwan, line 451: Add a missing space between `lumi_mask` and `=`.

cschwan:
> There are only a few things that are not exactly the same as before (and that we may or may not want to address):
> 
>     * the methods `.create()` and `.from_grid()` are no longer available, instead objects are instantiated from the class
> 
>     * `.into()` needs to be manually called on subgrids in order to convert into `SubgridEnum`

That's perfectly fine, and quite happy that we can get away with so few changes actually.

> 
> 
> PS: I could not force-push so please ignore the commit history 😇

I've disabled force-push for the time being since I'm not convinced about it being a good idea.

pineappl_py/src/grid.rs
felixhekhorn: I guess running `cargo fmt` might be enough? can we convince you to activate pre-commit :innocent: ? Look e.g. at [eko](https://github.com/NNPDF/eko/blob/b33509a03cd3ff5a03e6cb535da49617d84937bf/.pre-commit-config.yaml#L46)

felixhekhorn:
> * the methods `.create()` and `.from_grid()` are no longer available, instead objects are instantiated from the class

I think this is acceptable - but makes this a breaking change.

> * `.into()` needs to be manually called on subgrids in order to convert into `SubgridEnum`

this is quite "rusty" and not "pythonic", if we can't find a way around we can keep it to honour our underlying language - but if possible I'd like to avoid. Can we do something with the `Into` trait? i.e. implement that for the subgrids (i.e. move the function to an `impl` block) and then take the trait as argument of `PyGrid::set_subgrid`? (No idea if PyO3 will like it)

pineappl_py/src/grid.rs
felixhekhorn: `PyReadonlyArray1` should come from the [numpy create](https://docs.rs/numpy/latest/numpy/borrow/type.PyReadonlyArray1.html) and thus should correspond to a numpy array - instead `Vec` is `std` and thus should correspond to a (builtin) `list`

pineappl_py/src/bin.rs
felixhekhorn: since in Rust one can do `use pineappl::bin::BinRemapper as RustBinRemapper` I guess it should be possible. We were most likely influenced by the PyO3 naming scheme, but there is no need to follow them - things are scoped by crate. However, note that this might trigger more changes (in number of changed characters) than the present version.

pineappl_py/src/grid.rs
Radonirinaunimi: > I guess running `cargo fmt` might be enough? can we convince you to activate pre-commit 😇 ? Look e.g. at [eko](https://github.com/NNPDF/eko/blob/b33509a03cd3ff5a03e6cb535da49617d84937bf/.pre-commit-config.yaml#L46)

Would it? I have automatic formatting (which should be running `cargo fmt` under the hood) when saving on my helix editor but it wasn't able to pick that up.

pineappl_py/src/grid.rs
Radonirinaunimi: > `PyReadonlyArray1` should come from the [numpy create](https://docs.rs/numpy/latest/numpy/borrow/type.PyReadonlyArray1.html) and thus should correspond to a numpy array - instead `Vec` is `std` and thus should correspond to a (builtin) `list`

Yes, or also a `tuple`.

pineappl_py/src/bin.rs
Radonirinaunimi: I don't have a preference for this. Adding custom names (for what are exposed) required the lowest amount of changes while still complying with the PyO3 naming scheme, but if you really want to change these I think we can do so.

Radonirinaunimi: Actually, in hindsight, I think it is preferable to keep it this way - this is exactly the way PyO3 recommends.

pineappl_py/src/grid.rs
cschwan: It wouldn't, because `cargo fmt` won't change the contents of macros it doesn't know, because there whitespace may make a difference.

pineappl_py/src/grid.rs
cschwan: So which one is more generic? If we change it to a `Vec`, would numpy arrays still be accepted?

Radonirinaunimi: > So which one is more generic? If we change it to a Vec, would numpy arrays still be accepted?

`Vec<PyObject>` is more general than `PyReadonlyArray1<PyObject>`  as the former accepts all `list`, `tuple`, and `np.ndarray` while the latter can only accept `np.ndarray`. The exception is when `PyObject=bool`. This is because `bool` is not compatible with `np.bool_`.

As a specific exception, `Vec<(f64, f64)>` only accepts `tuple(float, float)` - which is what we want for `xi`.

Radonirinaunimi:
> this is quite "rusty" and not "pythonic", if we can't find a way around we can keep it to honour our underlying language - but if possible I'd like to avoid. Can we do something with the `Into` trait? i.e. implement that for the subgrids (i.e. move the function to an `impl` block) and then take the trait as argument of `PyGrid::set_subgrid`? (No idea if PyO3 will like it)

I've had a look at this a bit and not only it could become really complicated but also messy. What would be the reasons we should avoid the current approach (apart from breaking changes)? It is basically a difference between doing:
```python
set_subgrid(order, bin, lumi, subgrid.into())
```
vs before when it was re-wrapped inside:
https://github.com/NNPDF/pineappl/blob/8c1233504e06c80f5943013a5a4d79c3bb187d1e/pineappl_py/pineappl/grid.py#L127-L143

pineappl_py/src/grid.rs
cschwan: OK, thank you for looking this up! I presume it best we make as few changes as possible, so let's keep this in mind for the future.

cschwan:
> ```python
> set_subgrid(order, bin, lumi, subgrid.into())
> ```

Where do we use `Grid::set_subgrid`? This will need to changed, since soon (#299) all the subgrid types will be changed/removed.

felixhekhorn:
> Where do we use `Grid::set_subgrid`? This will need to changed, since soon (#299) all the subgrid types will be changed/removed.

[yadism](https://github.com/NNPDF/yadism/blob/2eab551ea5c9c8bea9c1695319a585ccf0986c5b/src/yadbox/export.py#L81) does - since we do the interpolation ourselves

> What would be the reasons we should avoid the current approach

as said above the only reason is being "rusty" and not "pythonic" - if it is too much of a hassle it is also fine ... (as Python knows no (strong) types, it also doesn't know casting so it feels a bit weird, that's it)

Radonirinaunimi:
> as said above the only reason is being "rusty" and not "pythonic" - if it is too much of a hassle it is also fine ... (as Python knows no (strong) types, it also doesn't know casting so it feels a bit weird, that's it)

Ok, let's keep it as it is then.

I think this is basically done. The remaining todos re docs and change of macros were already addressed by #301. The only maybe we'd add is a pre-commit hooks to format the python tests (and that also runs `cargo fmt`)(?).

pineappl_py/tests/test_lumi.py
felixhekhorn, line 1: remove this file - it is replaced by `test_channel.py` (which contains the same stupid stuff :see_no_evil: )

pineappl_py/src/evolution.rs
```diff
@@ -44,3 +44,9 @@ impl PyEvolveInfo {
 pub struct PyOperatorSliceInfo {
     pub(crate) slice_info: OperatorSliceInfo,
 }
+
+#[pymodule]
+pub fn evolution(_py: Python, m: &PyModule) -> PyResult<()> {
+    m.add_class::<PyEvolveInfo>()?;
```
felixhekhorn, line 50: - we need to expose `PyOperatorSliceInfo` for https://github.com/NNPDF/pineko/pull/181 
- although `pineappl::EvolveInfo` is not officially deprecated, I think it practice it is (i.e. I could not find a relevant use case) - so I suggest to drop it (even maybe in the main library)
- as we concluded that the naming scheme with `name` in the macro is the recommended way, we should propagate that through out - i.e. `PyOperatorSliceInfo` should be exposed as `OperatorSliceInfo` - in sum pineko should change from `from pineappl.grid import PyOperatorSliceInfo` to `from pineappl.evolution import OperatorSliceInfo`

felixhekhorn: wait a second - I think this module should just be dropped as is: `PyOperatorSliceInfo` is correctly exposed in `grid.rs`

pineappl_py/src/grid.rs
```diff
@@ -26,15 +26,15 @@ use pyo3::prelude::*;
...
 /// PyO3 wrapper to :rustdoc:`pineappl::grid::Order <grid/struct.Order.html>`.
-#[pyclass]
+#[pyclass(name = "Order")]
 #[repr(transparent)]
 pub struct PyOrder {
     pub(crate) order: Order,
 }
 
 // TODO: should probably be in a different module
 // TODO: rename to `PidBasis`
```
felixhekhorn, line 36:
```suggestion
```

pineappl_py/src/grid.rs
```diff
@@ -52,7 +52,7 @@ impl From<PyPidBasis> for PidBasis {
 
 // TODO: should probably be in a different module
 // TODO: rename to `OperatorSliceInfo`
```
felixhekhorn, line 54:
```suggestion
// TODO: should probably be in a different module
```
re other module - see my comment there, we could also shift - just decide something

pineappl_py/src/evolution.rs
Radonirinaunimi: Yes, `OperatorSliceInfo` is correctly exposed and `pineko` will be calling it as `from pineappl.grid import OperatorSliceInfo`.

> although pineappl::EvolveInfo is not officially deprecated, I think it practice it is (i.e. I could not find a relevant use case) - so I suggest to drop it (even maybe in the main library)

Just to confirm, you only want to remove the exposed `pineappl.evolution` module right? Because we cannot completely drop `evolution.rs` as it is used here:
https://github.com/NNPDF/pineappl/blob/93d3c634d8df92d6e310bc79d7d75e9e711d7054/pineappl_py/src/grid.rs#L607

and subsequently in `pineko`.

felixhekhorn: sorry, I was confused - of course `pineko` needs the module (I had the misconception that `EvolveInfo` and `OperatorSliceInfo` are the same - which of course they are not)

pineappl_py/src/grid.rs
felixhekhorn: shouldn't we move the class into the `evolution` module?

Radonirinaunimi: Yes, I think you are right. `OperatorSliceInfo` should be in the `evolution` module as it is a characteristic of the evolution. I will do so.

Radonirinaunimi: Please check if you are now fine with [fc6d466](https://github.com/NNPDF/pineappl/pull/302/commits/fc6d4667b36df601e850691b59b0569daa4d90b4).

cschwan (approved):

felixhekhorn (approved):

Radonirinaunimi:
Thanks @cschwan and @felixhekhorn! I will add a `CHANGELOG` later (do not have access to a computer right now) and then merge this.

CHANGELOG.md
```diff
@@ -12,6 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - added new method `Grid::delete_orders` and the corresponding switch
   `--delete-orders` in the subcommand `write` of the CLI
 
+### Changed
+
+- renamed `pineappl.lumi.rs` to `pineappl.channel.rs`
+- removed `.create()` and `.from_grid()` methods and directly instantiate objecs from the class
+- `.into()` needs to be explicitly called on subgrids when using `set_subgrid()`
+- replaced `pineappl.grid.PidBasis` with `pineappl.evolution.PidBasis`
+- replaced `pineappl.grid.OperatorSliceInfo` with `pineappl.evolution.OperatorSliceInfo`
```
felixhekhorn, line 21:
```suggestion
- replaced `pineappl.grid.PyPidBasis` with `pineappl.evolution.PidBasis`
- replaced `pineappl.grid.PyOperatorSliceInfo` with `pineappl.evolution.OperatorSliceInfo`
```

CHANGELOG.md
```diff
@@ -12,6 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - added new method `Grid::delete_orders` and the corresponding switch
   `--delete-orders` in the subcommand `write` of the CLI
 
+### Changed
+
+- renamed `pineappl.lumi.rs` to `pineappl.channel.rs`
+- removed `.create()` and `.from_grid()` methods and directly instantiate objecs from the class
```
cschwan, line 18: Removed items should be listed under a `### Removed` heading

CHANGELOG.md
```diff
@@ -12,6 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - added new method `Grid::delete_orders` and the corresponding switch
   `--delete-orders` in the subcommand `write` of the CLI
 
+### Changed
+
+- renamed `pineappl.lumi.rs` to `pineappl.channel.rs`
```
felixhekhorn, line 17:
```suggestion
- dropped top-level Python interface layer
- renamed 'lumi' to 'channel' in PyO3 Python interface. This concerns the 1) the argument names of `convolute_with_one` and similar functions; 2) the module `pineappl.lumi.rs` was moved to `pineappl.channel.rs`; 3) the class `LumiEntry` was renamed to `Channel`
```
a) let's be more explicit and b) we should add the changes from #301 as well

cschwan, line 17: prefix all of your items with `Python API:` to make clear that this concerns it and not the Rust API.

cschwan, line 17: This should probably talk about renaming the modules. What's unclear is that the first `.` refers to a module and the second one to file suffix.

felixhekhorn: I adjusted my suggested change to reflect that - do you agree with it?

CHANGELOG.md
felixhekhorn:
```suggestion
- removed `Grid.create()` and `FkTable.from_grid()` methods; use the constructors of the respective classes instead
```
+ I suggest a slight rewording

Radonirinaunimi:
Updated the logs to combine the comments/suggestions from review. Could you please check if these are fine?

CHANGELOG.md
felixhekhorn: @cschwan correct me if I'm wrong, but I believe what he meant is
```
- PythonAPI: dropped top-level Python interface layer
- PythonAPI: blub
```

CHANGELOG.md
```diff
@@ -12,6 +12,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - added new method `Grid::delete_orders` and the corresponding switch
   `--delete-orders` in the subcommand `write` of the CLI
 
+### Changed
+
+- dropped top-level Python interface layer
+- renamed `lumi` to `channel` in PyO3 Python interface. This concerns 1) the argument names of `convolute_with_one` and similar functions; 2) the module `Python API:pineappl.lumi` was moved to `Python API:pineappl.channel`; 3) the class `LumiEntry` was renamed to `Channel`
+- `.into()` needs to be explicitly called on subgrids when calling `Python API:pineappl.grid.set_subgrid()`
+- replaced `Python API:pineappl.grid.PyPidBasis` with `Python API:pineappl.evolution.PidBasis`
```
cschwan, line 20: sorry, what I meant was something along the lines of

- Python API: replaced `pineappl.grid.PyPidBasis` with `pineappl.evolution.PidBasis`

cschwan: Also manually break the lines after a maximum of 80 characters.

CHANGELOG.md
cschwan: ... with a space between 'Python' and 'API' though.

CHANGELOG.md
Radonirinaunimi: Also addressed in [230b188](https://github.com/NNPDF/pineappl/pull/302/commits/230b188c48df1d817e7dbf949f74d271356de50f).

cschwan (changes requested):
We have to fix the documentation generation, because currently it's broken. This concerns the files in `pineappl_py/docs/*`. It seems that most the stuff there is also probably outdated.

@alecandido you wrote most of the stuff in there, is this still relevant? In particular, do we need to concerns ourselves with any containers at this point?

I found this [blog post](https://blog.savant-ai.io/documenting-native-python-extensions-made-with-rust-and-pyo3-227aff68e481) that explains how to generate Sphinx documentation for the RTD, and its example files are here: <https://github.com/insight-platform/pyo3-sphinx-documentation>.

felixhekhorn:
I was already wondering if we might have a problem with documentation a few days ago and checked what we do and it seemed fine to me: specifically,
- we are using [the napoleon extension](https://github.com/NNPDF/pineappl/blob/70fa5a03de57c8409d84cc8e31d8d4111a690a33/pineappl_py/docs/source/conf.py#L45) as the blog post says
- we are using the compiled extension to generate the docs: https://github.com/NNPDF/pineappl/blob/70fa5a03de57c8409d84cc8e31d8d4111a690a33/pineappl_py/docs/source/conf.py#L166

> We have to fix the documentation generation, because currently it's broken.

where did you check? on RTD? what is the specific error?

alecandido:
> @alecandido you wrote most of the stuff in there, is this still relevant? In particular, do we need to concerns ourselves with any containers at this point?

No, definitely not.

PyO3 evolved a lot. And also my familiarity with it.
Having Python files was an easy way to keep using Rust inside, while not compromising the flexibility of the interface (NumPy arrays, default values, getters/setters, ...).
By now, PyO3 exposes all these features directly in Rust, so the Python files are just redundant. I stopped myself using them as much as possible (currently, the only use I'm making of them is incremental migration from a previous Python code base).

Concerning the docs, I believe there should be no problem at all: despite what it's being said in the blog post, the magic is all done by plain PyO3 and plain Sphinx.
Indeed, in Python functions are objects, and classes as well. So, PyO3 is creating instances of `PyObject` with the `.__doc__` attribute attached, which is standard in Python, to the point that is even part of the default parsers (when the first statement in a function, class, module is a string, that is attached to the object created as the `.__doc__` attribute). Sphinx is just reading these attributes.
About extensions:
- `-autodoc` is the one responsible to bridge the HTML generation from RST to the Python string (actually taking care of loading the `.__doc__` attributes, just importing the modules themselves, and scanning the reachable content)
- `-autosummary` is creating tables with the summary of `-autodoc` objects
- `-napoleon` is just a way to write nicer raw docstrings (with NumPy or Google format), instead of the default ones that are a bit crammed and verbose

Moreover, in principle you should write the directives for `-autodoc` manually, telling which class you want to document where.
However, this is already done automatically by another built-in tool, i.e. `-apidoc` (which is not properly an extension, but rather a command/application). You can see it here:
https://github.com/NNPDF/pineappl/blob/70fa5a03de57c8409d84cc8e31d8d4111a690a33/pineappl_py/docs/source/conf.py#L142-L170

In summary, it should already be done automatically, with nothing more needed on our side.

> We have to fix the documentation generation, because currently it's broken. This concerns the files in `pineappl_py/docs/*`. It seems that most the stuff there is also probably outdated.

It might well be, and even `conf.py` should be reviewed every now and then. However, I'd start from the `.rst` files.

alecandido:
> where did you check? on RTD? what is the specific error?

By the way, I'd suggest you to first attempt local compilation. It's always easier to debug.

If you fail to reproduce, then go for the CI :)

cschwan:
I updated the `README.md` with instructions on how to generate the documentation. However, then I get:

Screenshot_20240814_111709

So where the documentation should be, there's a warning/error.

felixhekhorn:
> where the former was the syntax that did not work.

using the hack [here](https://github.com/PyO3/pyo3/issues/1517#issuecomment-808664021) I can make `$ python -c "import pineappl.channel"` working again and maybe we want to do that; However, this still does not solve the problem of the submodules (or at least the ones I hacked) not to be listed ...

alecandido:
> using the hack [here](https://github.com/PyO3/pyo3/issues/1517#issuecomment-808664021) I can make `$ python -c "import pineappl.channel"` working again and maybe we want to do that; However, this still does not solve the problem of the submodules (or at least the ones I hacked) not to be listed ...

Submodules are complex beasts, not much because of PyO3 itself, but because of the way Python organizes the compiled extensions.

However, you can try to do something like:
```py
import pineappl.grid
import pineappl.channel
...
```
and so on directly at the beginning of `conf.py`, to make sure they are loaded when Sphinx will try to inspect them.

alecandido:
Try also manually with `-autodoc` what you can do, before turning to full automation with `-apidoc` (at least you can debug one problem at a time)

felixhekhorn:
> Try also manually with `-autodoc` what you can do, before turning to full automation with `-apidoc` (at least you can debug one problem at a time)

adding manually `.. automodule:: pineappl.grid` works just fine and I guess this would be our escape option - the problem is really `apidoc`

> However, you can try to do something like:

that doesn't help either unfortunately

> Submodules are complex beasts,

this is indeed the problem - because apidoc thinks here
https://github.com/sphinx-doc/sphinx/blob/fadb6b10cb15d2a8ce336cee307dcb3ff64680bd/sphinx/ext/apidoc.py#L344
pineappl is just a module - and I believe it does so, because it looks to the _file_ structure and not the ast ... I don't know if there is much that we can do at this point ... shall we take the manual route?

felixhekhorn:
> Would it simplify the situation if we put everything into a single module?

nop (as far as I understand)

> I've been thinking about doing the same at the Rust level, because submodule don't add any value IMO.

please don't - we want namespaces to organize things

> shall we take the manual route?

done in https://github.com/NNPDF/pineappl/pull/302/commits/144e7917fb2673d89052d7996e1a99a3cad5f94d for the moment, if someone has a better idea, please go ahead

Radonirinaunimi:
> Would it simplify the situation if we put everything into a single module? So instead of `pineappl.grid.Grid` it would be `pineappl.Grid`.
> 
> I've been thinking about doing the same at the Rust level, because submodule don't add any value IMO.

This was done here (https://github.com/NNPDF/pineappl/pull/302/commits/0a148089a0af0a9d5eeb306bab6116a2567f18a0) before I reverted it to the worry that it'd lead to too many breaking changes.

felixhekhorn:
> using the hack [here](https://github.com/PyO3/pyo3/issues/1517#issuecomment-808664021) I can make `$ python -c "import pineappl.channel"` working again and maybe we want to do that;

done in https://github.com/NNPDF/pineappl/pull/302/commits/8bce28c19f510c85331f04e9bd88a8bedbeeb3aa

felixhekhorn:
The current status of the docs is now available here: https://pineappl.readthedocs.io/en/remove_pymodules/index.html

pineappl_py/docs/.gitignore
```diff
@@ -1,9 +0,0 @@
-# autogenerated file
-source/development/code_todos.rst
-# Ignore auto generated module references
-source/modules
```
alecandido, line 4: I'd suggest adding this back, eventually

pineappl_py/docs/source/recipes.rst
```diff
@@ -13,7 +13,7 @@ How can I convolve a given PineAPPL grid with my PDF?
     import lhapdf
     g = pineappl.grid.Grid.read("path/to/grid.pineappl.lz4")
     pdf = lhapdf.mkPDF("YourPDF", 0)
-    bins = g.convolve(pdf.xfxQ2, pdf.xfxQ2, pdf.alphasQ2)
+    bins = g.convolve_with_one(2212, pdf.xfxQ2, pdf.alphasQ2)
```
felixhekhorn, line 16: We should make these snippets running at some point (not now) - e.g. turning the page into a notebook - to see if our examples actually work (which they were not before - I hope they do now :see_no_evil: ). Note the dependency on lhapdf, but as I found out out the other day in yadism - running the notebook locally should be sufficient for RTD

alecandido: There are doctests:
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

pineappl_py/docs/.gitignore
cschwan: Not the last line though, at least for now we manually created the file `pineappl_py/docs/source/modules/pineappl.rst`.

pineappl_py/docs/source/modules/pineappl.rst
```diff
@@ -0,0 +1,42 @@
+pineappl package
```
cschwan, line 1: Replace this with `The Python API of PineAPPL`.

alecandido:
> This was done here ([0a14808](https://github.com/NNPDF/pineappl/commit/0a148089a0af0a9d5eeb306bab6116a2567f18a0)) before I reverted it because of the worry that it'd lead to too many breaking changes.

@Radonirinaunimi, btw, you can even keep the two things at the same time https://github.com/PyO3/pyo3/issues/759#issuecomment-2282197848 (i.e. the workaround and the declarative module).

pineappl_py/requirements.txt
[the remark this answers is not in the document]
cschwan: This file really is for generating the documentation, so maybe it's better placed into `docs` (?). If that the better place, also correct `pineappl_py/README.md`.

alecandido:
And the tracking issue for `#[pymodule]` progresses is this one https://github.com/PyO3/pyo3/issues/3900

At the bottom, issues with docs are also mentioned.

pineappl_py/requirements.txt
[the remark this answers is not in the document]
felixhekhorn: done in [09460b6](https://github.com/NNPDF/pineappl/pull/302/commits/09460b6ec1f7502e1c72e75e0ffc1f4cd34b9af4)

pineappl_py/docs/source/modules/pineappl.rst
felixhekhorn: done in 53fba7ba

Radonirinaunimi:
> > This was done here ([0a14808](https://github.com/NNPDF/pineappl/commit/0a148089a0af0a9d5eeb306bab6116a2567f18a0)) before I reverted it because of the worry that it'd lead to too many breaking changes.
> 
> @Radonirinaunimi, btw, you can even keep the two things at the same time [PyO3/pyo3#759 (comment)](https://github.com/PyO3/pyo3/issues/759#issuecomment-2282197848) (i.e. the workaround and the declarative module).

Yes, seen this but did not directly resort to it as it requires the experimental feature.

pineappl_py/docs/.gitignore
felixhekhorn: done in [09460b6](https://github.com/NNPDF/pineappl/pull/302/commits/09460b6ec1f7502e1c72e75e0ffc1f4cd34b9af4)

cschwan:
> > @Radonirinaunimi, btw, you can even keep the two things at the same time [PyO3/pyo3#759 (comment)](https://github.com/PyO3/pyo3/issues/759#issuecomment-2282197848) (i.e. the workaround and the declarative module).
> 
> Yes, seen this but did not directly resort to it as it requires the experimental feature.

This feature has been stabilized in PyO3 0.22; we'd just need to upgrade from 0.21.

pineappl_py/requirements.txt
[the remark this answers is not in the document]
felixhekhorn: what @alecandido says is the correct file location is `pyproject.toml` - and we are indeed specifying the relevant packages and their versions there;
https://github.com/NNPDF/pineappl/blob/53fba7ba5354fe7179f6969f896329ea00efecc0/pineappl_py/pyproject.toml#L28-L33

so why do we actually need `requirements.txt` in the first place?

felixhekhorn:
> This feature has been stabilized in PyO3 0.22; we'd just need to upgrade from 0.21.

please let's not do this in this PR - else we never merge :see_no_evil:

pineappl_py/docs/.gitignore
cschwan, line 1: Please use `pineappl_py/.gitignore`, we shouldn't nest them too deeply.

pineappl_py/requirements.txt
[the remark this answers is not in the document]
cschwan: Using `pyproject.toml` seems like the better solution. Let's use it instead of `requirements.txt`.

pineappl_py/docs/.gitignore
felixhekhorn: done as requested in 2727c4cb - but I would have said the other way round: make them as local (i.e. as deep) as possible

pineappl_py/docs/source/conf.py
```diff
@@ -1,174 +1,65 @@
...
+html_theme = 'sphinx_rtd_theme'
+html_static_path = ['_static']
 
-# Add any paths that contain custom static files (such as style sheets) here,
-# relative to this directory. They are copied after the builtin static files,
-# so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ["_static"]
 
-# -- Extension configuration -------------------------------------------------
+here = pathlib.Path(__file__).absolute().parent
```
cschwan, line 46: This line should also be commented out if it's not used.

pineappl_py/.gitignore
```diff
@@ -21,8 +21,13 @@ coverage.xml
...
 # Sphinx documentation
-docs/_build/
-doc/_build/
+# Ignore auto generated module references
+# TODO: while apidoc does not work, we need to preserve that folder
+# docs/source/modules
+# ignore temporary build files
+docs/build/
+# Ignore generated sphinx-bibtex file
+docs/source/bibtex.json
```
cschwan, line 30: We don't use bibtex (and I don't think we ever did), so let's remove this line.

pineappl_py/docs/source/conf.py
felixhekhorn: done in 77995e07

pineappl_py/.gitignore
felixhekhorn: done in 77995e07 (as you can guess from earlier this is just copied from eko)

pineappl_py/requirements.txt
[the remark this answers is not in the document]
felixhekhorn: done in [19e1564](https://github.com/NNPDF/pineappl/pull/302/commits/19e15646962c97244f1e5d4b7446a60e1fcce113)

felixhekhorn (approved):
I didn't check everything again, but just skimmed over and it seems alright ...

pineappl_py/README.md
```diff
@@ -19,11 +19,19 @@ Run
 
 ```shell
 python -m venv env && . env/bin/activate
-pip install -r requirements.txt
+pip install .[docs]
```
cschwan, line 22: That doesn't seem to work.

Radonirinaunimi: Maybe you need quotes? Ie `pip install '.[docs]'`.

cschwan: No, even `pip install .` doesn't work. Should be fixed with commit 6895357718ef53ccda0fea63d0fa9b87ffcb0329.

pineappl_py/pyproject.toml
```diff
@@ -28,7 +28,6 @@ cli = ["pineappl-cli"]
 docs = [
   "sphinx>=6.2.1",
   "sphinx_rtd_theme>=1.2.2",
-  "sphinxcontrib-bibtex>=2.5.0",
   "nbsphinx>=0.9.2",
```
felixhekhorn, line 31:
```suggestion
```

pineappl_py/README.md
felixhekhorn: Mmm this is precisely what RTD is doing: https://readthedocs.org/projects/pineappl/builds/25310692/

cschwan: I'm getting:

```
Processing /home/cschwan/projects/pineappl/pineappl_py                                                                                        
  Installing build dependencies ... done 
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error                                 
    ERROR: Command errored out with exit status 1:
     command: /home/cschwan/projects/pineappl/pineappl_py/env/bin/python3 /tmp/tmptb0qt4nl_in_process.py prepare_metadata_for_build_wheel /tmp/tmpl83sw3iv
         cwd: /tmp/pip-req-build-55h9jzl7                                                                                                     
    Complete output (13 lines):      
    error: failed to parse manifest at `/tmp/pip-req-build-55h9jzl7/Cargo.toml`
                                                                                                                                              
    Caused by:                      
      error inheriting `version` from workspace root manifest's `workspace.package.version`
                                            
    Caused by:                                                                                                                                
      failed to find a workspace root                                                                                                         
    💥 maturin failed
      Caused by: Cargo metadata failed. Does your crate compile with `cargo build`?
      Caused by: `cargo metadata` exited with an error:                
    Error running maturin: Command '['maturin', 'pep517', 'write-dist-info', '--metadata-directory', '/tmp/pip-modern-metadata-2wy79mhg', '--interpreter', '/home/cschwan/projects/pineappl/pineappl_py/env/bin/python3']' returned non-zero exit status 1.
    Checking for Rust toolchain....      
    Running `maturin pep517 write-dist-info --metadata-directory /tmp/pip-modern-metadata-2wy79mhg --interpreter /home/cschwan/projects/pineappl/pineappl_py/env/bin/python3`
    ----------------------------------------
WARNING: Discarding file:///home/cschwan/projects/pineappl/pineappl_py. Command errored out with exit status 1: /home/cschwan/projects/pineappl/pineappl_py/env/bin/python3 /tmp/tmptb0qt4nl_in_process.py prepare_metadata_for_build_wheel /tmp/tmpl83sw3iv Check the logs for full command
 output.      
ERROR: Command errored out with exit status 1: /home/cschwan/projects/pineappl/pineappl_py/env/bin/python3 /tmp/tmptb0qt4nl_in_process.py prepare_metadata_for_build_wheel /tmp/tmpl83sw3iv Check the logs for full command output.
```

felixhekhorn: yes, I'm getting the same (I already forgot) - I patched the `Cargo.toml` :see_no_evil: ; but so it is not only me ... however this sounds like a maturin problem maybe ... (and RTD can do it)

felixhekhorn: > or actually the manifest definition,

he doesn't like the "inherit from parent package" - if you copy then plain version from above (what my hack is) is works

alecandido: In any case, the fix in the README is correct: if you want to use `maturin develop` you need `pip install maturin` first.

Otherwise, you can use:
```sh
pip install .     # install from source
pip install -e .  # install from source in editable mode
```
and the second is equivalent to `maturin develop`. But it's an alternative, downloading and using Maturin internally. You should choose your workflow.

alecandido: > he doesn't like the "inherit from parent package" - if you copy then plain version from above (what my hack is) is works

And trying with Maturin 1.7?
The version written here is a lower bound, but dating almost two years ago (with a 1.0 release in between...).

cschwan: For me it doesn't - I've always used 1.7.

cschwan: This was caused by an outdated version of `pip`. I updated our README in commit b0de152dfe81591158f25b3b1d6602c150dfc919, which now describes the potential pitfall.

cschwan (approved):

pineappl_py/README.md
```diff
@@ -12,3 +12,49 @@ For installation instructions see the [documentation].
...
+
+to build the project, which also installs it into the environment so that it
+can be used in Python projects that use the same environment.
+
+### Documentation
+
+Run the following once to install the documentation's dependencies:
+
+```shell
+pip install .[docs]
```
alecandido, line 51: I'm not sure whether I'd recommend single quotes:
```suggestion
pip install '.[docs]'
```

I know that it is not required with Bash and Nu, but it is required with Zsh, which is the default on MacOS.
Of course, I hope people using Zsh are already aware of it. But when it's a combination of the default shell and a non-ubiquitous feature (extras), you know someone may experience the issue. Though, it could be easily solved.

The single quotes instead are always working in my experience (despite being less elegant and a bit more annoying...).

Radonirinaunimi:
Are we happy to merge this now? We can add pre-commit hooks to automatically run black and cargo fmt before commit in another PR.

cschwan:
> Are we happy to merge this now?

I found some minor things, let me fix them first.

Radonirinaunimi:
@cschwan What was actually the reason of renaming/moving `channel` into `boc`?

cschwan:
> @cschwan What was actually the reason of renaming/moving `channel` into `boc`?

That just reflects the same module structure in Rust. I think we should keep the two as close as possible.

Radonirinaunimi:
> That just reflects the same module structure in Rust. I think we should keep the two as close as possible.

Yes, this I understood and I completely agree. But what was the reason to change it in the main Rust module (for my own curiosity).

cschwan:
> But what was the reason to change it in the main Rust module (for my own curiosity).

`Grid` is basically, as you know, a three-dimensional array over bins, orders and channels, and I wanted to have a single module to reflect these three dimensions. It's not perfect yet, as there's still the `bin` module and the ordering isn't `boc` but `obc`, but I'd expect that to change in the future.

Radonirinaunimi:
> `Grid` is basically, as you know, a three-dimensional array over bins, orders and channels, and I wanted to have a single module to reflect these three dimensions. It's not perfect yet, as there's still the `bin` module and the ordering isn't `boc` but `obc`, but I'd expect that to change in the future.

I see, thanks!

cschwan:
@Radonirinaunimi can you please check that commit 8ef229741373a9a49fb67883f69e228de8d5a66d doesn't do anything unexpected? Contrary to what I wrote it isn't a bug, in the new commit it should just be more succinct. If it's OK, you can merge!

CHANGELOG.md
```diff
@@ -12,6 +12,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - added new method `Grid::delete_orders` and the corresponding switch
   `--delete-orders` in the subcommand `write` of the CLI
 
+### Changed
+
+- Python API: dropped top-level Python interface layer
+- Python API: renamed `lumi` to `channel` in PyO3 Python interface. This
+  concerns 1) the argument names of `convolute_with_one` and similar functions;
+  2) the module `pineappl.lumi` was moved to `pineappl.channel`; 3) the class
```
felixhekhorn, line 20:
```suggestion
  2) the module `pineappl.lumi` was moved to `pineappl.boc`; 3) the class
```
should we add more details to the changelog?

Radonirinaunimi: Yes, that one should definitely be fixed. Maybe @cschwan wants to add more information there?

Radonirinaunimi:
> @Radonirinaunimi can you please check that commit [8ef2297](https://github.com/NNPDF/pineappl/commit/8ef229741373a9a49fb67883f69e228de8d5a66d) doesn't do anything unexpected? Contrary to what I wrote it isn't a bug, in the new commit it should just be more succinct. If it's OK, you can merge!

I checked and this does not change anything.

CHANGELOG.md
cschwan: Good catch!

PR #305, opened 2024-08-30
v0.7.x

cschwan:


felixhekhorn:
let me repeat again what I was asking in #308 

---

also note that in v0.7.4 the error is slightly different:
```
return self.raw.convolute_with_two(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument 'bin_indices': type mismatch:
 from=int64, to=uint64
```
do we want to keep fixing former version, or not?

cschwan:
I'm not sure I understand this: is 0.7.x broken in any way?

felixhekhorn:
> I'm not sure I understand this: is 0.7.x broken in any way?

the `bin_indices` argument of `convolute_with_*` is (or seems to be) broken - but broken in a different way then master #308

cschwan:
I wouldn't say it's broken, it's just (from the POV of Python) unexpectedly strict. You must pass a numpy array and the types must be unsigned integers. I suppose that really is unnecessarily strict, and we should allow more flexibility. However, this would probably be a breaking change, so let's not change anything for 0.7.x.

cschwan:
This branch is outdated; before we make another `0.7.x` release, we should first merge the following commits (probably):

1. e6a2dea855a10c9ef7cd7868815758a9c4b261af
2. 49e1c08609be2716538388e6c1e0e79e52a10743
3. a3a080477c38fe58fa3fb70029b5e4476be05244
4. 76a1e89153c49acc7b0f2682b2d9ed05b145130a

PR #309, opened 2024-09-18
Replace recipes with jupyter notebook

Radonirinaunimi:
This is something I had started a while ago but hasn't pushed as I believe the tutorial could be a bit more extensive but decided to do so anyway in case you have suggestions on which aspects to cover.


felixhekhorn (commented):
Unfortunately I can not comment on the notebook lines, so I just collect some points here.
- this is already better than before, but since we are already on it let's make it even better
- I would consider introducing `pandas` - I find it quite useful for this kind of problems; (see also Como :upside_down_face: )
- don't print `bin_left` and `bin_right` separately but zip them (e.g. this would be more natural in pandas)
- we need to explain channels and orders
- maybe we want to give some sort of physics motivation, why one would adjust the bin normalization (e.g. plotting reasons)

Radonirinaunimi:
> Unfortunately I can not comment on the notebook lines, so I just collect some points here.
> 
>     * this is already better than before, but since we are already on it let's make it even better
> 
>     * I would consider introducing `pandas` - I find it quite useful for this kind of problems; (see also Como 🙃 )
> 
>     * don't print `bin_left` and `bin_right` separately but zip them (e.g. this would be more natural in pandas)
> 
>     * we need to explain channels and orders
> 
>     * maybe we want to give some sort of physics motivation, why one would adjust the bin normalization (e.g. plotting reasons)

Great, thanks! I will address these. These are mainly points that concern the current existing tutorials, are there some PineAPPL concepts that should be definitely highlighted on top of these (combining grids, etc.)?

felixhekhorn:
Mmm depends on how much we want to stuff into this ... if we want to add more, then I guess that one of the first things people want to know is more about `convolve_with_one`: all arguments are useful for physics, plotting, numerics. Also the `scale*` methods are sometimes useful ...

One thing that we should maybe link is the (Python) example in the repo - and maybe we want to even show it verbatim in the docs (next to recipes e.g.).

Radonirinaunimi:
Ok, I believe the tutorial is now complete. It includes both basic and (slightly) advanced (based on the `examples` use cases.

felixhekhorn:
> One thing that we should maybe link is the (Python) example in the repo - and maybe we want to even show it verbatim in the docs (next to recipes e.g.).

Mmm now we have the Python tutorial _almost the same_ twice, I guess we want to maintain only one ... so do we want to drop one, e.g. the one in examples and just point to the documentation there instead? Can we enforce the example being run also here? i.e. execute the notebook upon release? which I think is not happening by default

Radonirinaunimi:
So I removed the `examples/python/dyaa.py` as it is a bit of duplication (and it was deprecated).

> Can we enforce the example being run also here? i.e. execute the notebook upon release? which I think is not happening by default

In principle we can run all of the examples in some CI, at least to make sure that they are always up to date with the changes. But I'd do so in a separate PR.

Also, btw, it would be nice to to have this up online before the code meeting.

cschwan (approved):

Issue #180, opened 2022-09-21
Add pre-commit?


felixhekhorn:
How about activating [pre-commit](https://pre-commit.com/) here? e.g. [this repo](https://github.com/doublify/pre-commit-rust/blob/master/.pre-commit-hooks.yaml) provides `fmt`, `check` and  `clippy`

felixhekhorn:
I think we would only need to update if `cargo` would change, which is highly unlikely - in any case the hooks are very simple so we could even fork and maintain them ...

cschwan:
I agree that using a git hook is potentially a good idea. Right now we only write what people should do in [`CONTRIBUTING.md`](https://github.com/N3PDF/pineappl/blob/master/CONTRIBUTING.md), with hooks we could automate/enforce it. However, it seems that [pre-commit](https://pre-commit.com/) is mostly tailored towards Python and probably a bit too much for what we need; let's talk about that first.

- Run `cargo fmt --check` to see if everything is properly formatted.
- We probably don't want to run `cargo clippy` now, as this returns a very long list of warnings. At some point, however, we might consider doing that.
- Do we want to run `cargo check`? If yes, do we want to run it with `--features=applgrid,fktable,fastnlo` or a subset of these (which would require having these libraries installed)?
- I don't think we want to run `cargo test`, which is a job for our CI.

An alternative to [pre-commit](https://pre-commit.com/) is writing the hooks manually, putting them into `.githooks/` and activating them with `git config core.hooksPath .githooks/`.

AleCandido:
You are definitely right that `pre-commit` is coming from Python ecosystem, but in a sense is only the driver. Could have been written in shell or perl, and it was the same.

The main argument in favor of `pre-commit` is that simplifies a set of operations, and provide useful hooks out of the box. The main con is that it is definitely "non-minimal".

The minimal solution would be `.githooks/`, but we are adding too much to it, at some we'll start rewriting `pre-commit`. At that point the minimality would be defeated by duplication.

Issue #249, opened 2023-10-24
Improve `pineappl plot` Python interaction


alecandido:
Currently, `pineappl plot` relies on Python and Matplotlib. This is actually quite nice, because people using PineAPPL are often familiar with Python, and know how to plot the information, once given.

A current mild pain-point is that there is a lot of string manipulation involved in the call, since a Python script is generated everytime, embedding the data:
https://github.com/NNPDF/pineappl/blob/3476629d40f4b4af164ce067ec2d029781be048c/pineappl_cli/src/plot.rs#L508-L521

However, dumping the script is a design featured, in such a way that people are able to customize it.

I see a few possible ways of improving:
1. split the two `pineappl plot` operations: dump the script and the data separately (in a common data format, like CSV)
2. leverage `pineappl_py` in the script, and directly extract the data from the grid (requires the installation of the PineAPPL python extension)
3. move the plotting feature to `pineappl_py` itself (possibly with a `pineappl-plot` CLI, but it would be installed separately, at least until #176 will be completed)
4. use PyO3 to make the call from the CLI (i.e. from Rust) - in this way people will be able to customize the script and pass it again, but there won't be any need of dumping the data any longer (most likely this is an overkill)

AleCandido:
This is not urgent at all, it is just a consideration that came to mind reading #248, nothing more.

cschwan:
I had similar considerations in the past, but
1. I like having the data and the script in one file, because then I can easily move/copy/rename the single file. This will become more complicated if we split things into more files.
2. Using `pineappl_py` in the Python plotting script is possible, but then we introduce two more dependencies: pineappl_py and LHAPDF. Especially the latter, particularly the Python version, is a pain in the neck to get working. Another downside is that the script will take much longer, which we probably want to avoid, given that we often work on the Python script to modify the plot, without having to regenerate the data.
3. Having the functionality available from Python we should definitely consider, although my plan was actually to eventually expose the entire CLI functionality in another, higher-level, crate, for which we then can write Python bindings, of course.
4. This suggestion I don't understand.

That being said I believe the plotting script(s) and functionality can be significantly improved. I was thinking of using a templating engine in such a way that the template is really kept in one file. This mostly is the case right now (in this file: <https://github.com/NNPDF/pineappl/blob/master/pineappl_cli/src/plot.py>), but writing out the data could be improved, for instance.

AleCandido:
[...]

PR #310, opened 2024-09-18
Introduce pre-commit hooks

Radonirinaunimi:
My initial idea was to introduce pre-commit hooks to lint and format the python files, but then I stumbled upon https://github.com/NNPDF/pineappl/issues/180 and decided to introduce some minimal hooks for Rust and all the other files.

The configuration is kept to be very minimal in order to only perform the most necessary steps before pushing commits.

Left to do:
- ~[ ] fixing `pineappl_cli/src/plot.py`~ Addressed in https://github.com/NNPDF/pineappl/pull/312


.pre-commit-config.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - repo: local
+    hooks:
+    - id: fmt
+      name: cargo fmt
+      description: Format Rust files with cargo fmt.
+      entry: cargo fmt --
+      language: system
+      files: ^pineappl\S*\/.*\.rs$
+      args: []
+    - id: check
```
felixhekhorn, line 30: In eko we have `cargo clippy` but not `cargo check` - which one is better? will either `fmt` or `clippy` run `check` along the way?

.pre-commit-config.yaml
felixhekhorn, line 1: In MaunaKea I also have `clang-format` https://github.com/felixhekhorn/MaunaKea/blob/d857a363719a9e6f579f156948b483e4cbde3976/.pre-commit-config.yaml#L13

I also have `cpplint` but that is maybe not installed by default, unlike `clang-format` which might be available on most systems ...

felixhekhorn (approved):
I like it if someone else is doing the trivial stuff for me :upside_down_face:

.pre-commit-config.yaml
alecandido: They are different: `cargo check` is essentially running the compiler, without generating code. `cargo clippy` is statically analyzing the code (i.e. the equivalent of Pylint).

There is a non-negligible change that `cargo clippy` is a superset of `cargo check`, but it's also possible that they only partially overlap. But this you can search for (or even try: run `cargo clippy` on something that `cargo check` was flagging, and see if it's complaining - even a missing `;` or unmatched `{`).
The opposite is definitely impossible: `cargo clippy` may do strictly *more*, but certainly not strictly *less*.

alecandido:
> I like it if someone else is doing the trivial stuff for me 🙃

If it's trivial, it would take no time to you as well :P

I.e. use the word *trivial* with a large amount of caution

.pre-commit-config.yaml
Radonirinaunimi: > They are different: cargo check is essentially running the compiler, without generating code. cargo clippy is statically analyzing the code (i.e. the equivalent of Pylint).

Yes, exactly. I did not include `clippy` before because it is very noisy but there is actually some ways to suppress warnings and only show errors.

Radonirinaunimi:
It seems trivial indeed but I had to fight with python files which do not actually contain "python codes" (which in the end is time-consuming) 😇.

Btw, @cschwan, is there a reason why [pineappl_cli/src/plot.py](https://github.com/NNPDF/pineappl/blob/master/pineappl_cli/src/plot.py) can't just be modified into `plot.txt` or something?

felixhekhorn:
> Btw, @cschwan, is there a reason why [pineappl_cli/src/plot.py](https://github.com/NNPDF/pineappl/blob/master/pineappl_cli/src/plot.py) can't just be modified into `plot.txt` or something?

for sure we want the Python syntax highlight, because it is almost a Python file (so I'd prefer the `.py`).

However, I would much prefer making this a _true_ Python script as discussed in https://github.com/NNPDF/pineappl/issues/249 and split the static stuff in a true Python file and then all the dynamic stuff can either be inlined (in Rust) or could then go to a `config_plot.py.txt`

cschwan:
I thought about improving the plot script, but it didn't go anywhere. You're welcome to try to improve it yourself. The only conditions for improving it are:

1. the only dependencies you can rely on are: `numpy`, `matplotlib` and the ones already provided by Python.
2. everything has to be in one file.

.pre-commit-config.yaml
alecandido: Well: that's certainly an option, but another one is just implementing its suggestions :P

Radonirinaunimi: Yes, I agree! I will address this soon https://github.com/NNPDF/pineappl/issues/314 😉

cschwan:
@Radonirinaunimi please don't fix clippy's warnings, this is out of scope of this PR. Also don't silence them, because at some point we really want to fix them. For the time being we should limit the changes in `master` because the main development is going on in `v1-file-format`, which will change most of the code in the `pineappl` crate.

Radonirinaunimi:
> @Radonirinaunimi please don't fix clippy's warnings, this is out of scope of this PR. Also don't silence them, because at some point we really want to fix them. For the time being we should limit the changes in `master` to bugfixes and build-system related issues (like this PR) because the main development is going on in `v1-file-format`, which will change most of the code in the `pineappl` crate.

Agreed with limiting the scope of this PR to these. I will now revert the last commit. 

But as a side note, fixing all the clippy warnings, as was done in [9fbb246](https://github.com/NNPDF/pineappl/pull/310/commits/9fbb2464f0f9540e05e1420d631c602848e2091d), introduces not that many changes (as opposed to what I initially expected). The issue is that doing so changes the numerical values in a few places, which might not be worrying but for some lead the tests to fail.

cschwan:
> The issue is that doing so changes the numerical values in a few places, which might not be worrying but for some lead the tests to fail.

That's the problem I feared would appear. The evolution code, for instance, is quite fragile when it comes to numerical tolerances and will lead failures in unexpected places.

For the future: keep in mind that we will always want to prefer [`#[expect(...)]`](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#expectlint) instead of `#[allow(...)]`. That's a brand new feature, but starting with `v1` we'll be able to raise the MSRV. The warnings of the type `item in documentation is missing backticks` is a false-positive, and they're fixed by adding a `clippy.toml` (see commit 88f9151b96327c59760abc8c5ec7ff8cdf020be2).

Radonirinaunimi:
> For the future: keep in mind that we will always want to prefer [`#[expect(...)]`](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#expectlint) instead of `#[allow(...)]`. That's a brand new feature, but starting with `v1` we'll be able to raise the MSRV. The warnings of the type `item in documentation is missing backticks` is a false-positive, and they're fixed by adding a `clippy.toml` (see commit [88f9151](https://github.com/NNPDF/pineappl/commit/88f9151b96327c59760abc8c5ec7ff8cdf020be2)).

Thanks! That's very useful to know.

.pre-commit-config.yaml
```diff
@@ -5,7 +5,9 @@ repos:
     rev: v4.6.0
     hooks:
       - id: trailing-whitespace
-        exclude: ^pineappl_cli/tests/
+        # `.rs` files are taken care of by `cargo fmt`
+        # `.github/workflows/release.yml` contains vital spaces
```
felixhekhorn, line 9: shouldn't we rather move that part out of the `.yml` file?

Actually, having looked at the file, it is much too complicated I'd say - can't we split it into smaller snippets and scripts which are executed in the correct order inside the workflow?

cschwan (approved):

Radonirinaunimi:
Thanks @cschwan for fixing various details and polishing this. I guess this also can be merged?

cschwan:
Yes, I'll let you have the honours :smiley:!

.pre-commit-config.yaml
```diff
@@ -0,0 +1,51 @@
...
+    # A fast Python linter and code formatter. See
+    # https://docs.astral.sh/ruff/ for more details.
+    rev: v0.6.5
+    hooks:
+      - id: ruff
+        args: [--fix]
+        exclude: ^pineappl_cli/src/plot.py
+      - id: ruff-format
+        args: []
+        # TODO: remove this exclusion once we've merged this into master
```
felixhekhorn, line 22: Please remember this :upside_down_face:

Radonirinaunimi: You're right... This can be removed in master (I checked that the `plot.py` is not modified as it is already properly formatted).

cschwan: Done in commit cbe04a44fc3f0fcf87aa85ce3aec397eab9898ee.

PR #312, opened 2024-09-19
Make CLI plot script a proper Python file

### LINKED ISSUE #249: Improve `pineappl plot` Python interaction

felixhekhorn:
Closes #249 

given the constraints here https://github.com/NNPDF/pineappl/pull/310#issuecomment-2360684605, this is the best we can do - however, I still believe this is a step forward, because at least `plot.py` is now a genuine Python file. It also needs no longer (Rust) formatting, but just a plain find and replace.

@cschwan since you may know the answer straight away (in contrast to me): can you help me with:
https://github.com/NNPDF/pineappl/blob/300f3c65a12c973f9f019c707c9c889dfc606b85/pineappl_cli/src/plot.rs#L549-L550

also note that the plot script itself is badly documented, i.e. there is no documentation, just some matplotlib magic going on, which is hard to understand

I also ran `ruff` on it, which generates more changes

### DIFF
--- a/pineappl_cli/src/plot.py
+++ b/pineappl_cli/src/plot.py
@@ -188,14 +188,13 @@ def set_ylim(axis, save, load, filename):
     axis.set_ylim((ymin - space, ymax + space))
 
 
-def plot_int(axis, **kwargs):
+def plot_int(axis, /, pdf_results, **_kwargs):
     xmin = np.array([])
     xmax = np.array([])
     x = np.array([])
     y = np.array([])
 
-    for index, i in enumerate(kwargs["pdf_results"]):
-        label, ycentral, ymin, ymax = i
+    for index, (label, ycentral, ymin, ymax) in enumerate(pdf_results):
         x = np.append(x, ycentral[:-1])
         xmin = np.append(xmin, ymin[:-1])
         xmax = np.append(xmax, ymax[:-1])
@@ -221,16 +220,13 @@ def plot_int(axis, **kwargs):
     axis.margins(x=0.1, y=0.1)
 
 
-def plot_abs(axis, **kwargs):
-    x = kwargs["x"]
-    slice_label = kwargs["slice_label"]
-
+def plot_abs(axis, /, x, y, ymin, ymax, slice_label="", **_kwargs):
     axis.set_yscale("log" if ylog else "linear")
-    axis.step(x, kwargs["y"], colors[0], linewidth=1.0, where="post", label=slice_label)
+    axis.step(x, y, colors[0], linewidth=1.0, where="post", label=slice_label)
     axis.fill_between(
         x,
-        kwargs["ymin"],
-        kwargs["ymax"],
+        ymin,
+        ymax,
         alpha=0.4,
         color=colors[0],
         linewidth=0.5,
@@ -242,18 +238,14 @@ def plot_abs(axis, **kwargs):
         axis.legend()
 
 
-def plot_ratio_pdf(axis, **kwargs):
-    x = kwargs["x"]
-    slice_label = kwargs["slice_label"]
-    pdf_uncertainties = kwargs["pdf_results"]
+def plot_ratio_pdf(axis, /, x, pdf_results, slice_label="", **_kwargs):
+    axis.set_ylabel(ylabel_ratio_pdf.format(central_pdf=pdf_results[0][0]))
 
-    axis.set_ylabel(ylabel_ratio_pdf.format(central_pdf=pdf_uncertainties[0][0]))
-
-    for index, i in enumerate(pdf_uncertainties):
+    for index, i in enumerate(pdf_results):
         label, y, ymin, ymax = i
-        y = y / pdf_uncertainties[0][1]
-        ymin = ymin / pdf_uncertainties[0][1]
-        ymax = ymax / pdf_uncertainties[0][1]
+        y = y / pdf_results[0][1]
+        ymin = ymin / pdf_results[0][1]
+        ymax = ymax / pdf_results[0][1]
 
         axis.step(x, y, color=colors[index], linewidth=1.0, where="post")
         axis.fill_between(
@@ -272,7 +264,7 @@ def plot_ratio_pdf(axis, **kwargs):
         loc="upper left",
         mode="expand",
         borderaxespad=0,
-        ncol=min(4, len(pdf_uncertainties)),
+        ncol=min(4, len(pdf_results)),
     )
 
     if slice_label != "":
@@ -295,23 +287,19 @@ def plot_ratio_pdf(axis, **kwargs):
         )
 
 
-def plot_double_ratio_pdf(axis, **kwargs):
-    x = kwargs["x"]
-    slice_label = kwargs["slice_label"]
-    pdf_uncertainties = kwargs["pdf_results"]
-
+def plot_double_ratio_pdf(axis, /, x, pdf_results, slice_label="", **_kwargs):
     axis.set_ylabel(ylabel_double_ratio_pdf)
 
-    for index, i in enumerate(pdf_uncertainties):
+    for index, i in enumerate(pdf_results):
         label, y, ymin, ymax = i
         if index == 0 or index == 2:
-            y = y / pdf_uncertainties[0][1]
-            ymin = ymin / pdf_uncertainties[0][1]
-            ymax = ymax / pdf_uncertainties[0][1]
+            y = y / pdf_results[0][1]
+            ymin = ymin / pdf_results[0][1]
+            ymax = ymax / pdf_results[0][1]
         else:
-            y = y / pdf_uncertainties[1][1]
-            ymin = ymin / pdf_uncertainties[1][1]
-            ymax = ymax / pdf_uncertainties[1][1]
+            y = y / pdf_results[1][1]
+            ymin = ymin / pdf_results[1][1]
+            ymax = ymax / pdf_results[1][1]
         axis.step(x, y, color=colors[index], linewidth=1.0, where="post")
         axis.fill_between(
             x,
@@ -329,7 +317,7 @@ def plot_double_ratio_pdf(axis, **kwargs):
         loc="upper left",
         mode="expand",
         borderaxespad=0,
-        ncol=min(4, len(pdf_uncertainties)),
+        ncol=min(4, len(pdf_results)),
     )
 
     if slice_label != "":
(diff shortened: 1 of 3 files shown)


felixhekhorn:
Actually, after a bit more thinking I believe we can do even more, i.e. almost what I want:
1. I suggest to split all the static stuff, i.e. the plot functions, away from `pineappl_cli/src/plot.py` into a file `pineappl_py/something/plot.py`
2. that new file should be statically included into the `pineappl` Python package (and thus the Python interface will have easy access to the plot functions; we can also provide some additional methods which generate the necessary data from a grid using the Python interface itself)
3. the CLI will copy that file verbatim into the output of `pineappl plot`

pineappl_cli/src/plot.py
```diff
@@ -1,18 +1,40 @@
 #!/usr/bin/env python3
 
-import math
 import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
 import pickle
 
+# defaults for global variables coming from the CLI
+title = ""
```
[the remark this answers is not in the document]
felixhekhorn: done in 51a5d536

pineappl_cli/src/plot.rs
```diff
@@ -489,46 +506,84 @@ impl Subcommand for Opts {
...
+            let data = format!(
+                "data = {data_string}
+metadata = {{
+    {metadata}
+}}",
+                data_string = data_string,
                 metadata = format_metadata(&vector),
             );
+            let template = include_str!("plot.py");
+            // UNWRAP: if there are no markers the template is wrong
```
felixhekhorn, line 567: I'm almost sure this is not the most efficient way to do things, but it is the only one I managed to make running ...

alecandido: Well, it's not completely bad: you're distributing it within the binary.

It's true that in some other respects it may be better to install it somewhere on the file system, and recover it at runtime. But in this way you do not depend at all on the file system and anything else, and the price is increasing the binary size of something that is most likely negligible...
(it will be 10-20KB?)

felixhekhorn: nono, I don't mean the idea - it is the best we can do given the constraint (I think) and I will try to implement my idea [from above](https://github.com/NNPDF/pineappl/pull/312#issuecomment-2363034277) to keep the generated file simple and self-contained.

I was more worried of the string manipulation I'm doing here ... I was fighting with `split` and other stuff which did not work

cschwan:
There are a bit too many changes doing different things in here, which make reviewing and fixing the tests difficult. Could you please focus first on one goal, namely making `plot.py` a proper Python file? We can reformat and lint later on when we know it's working. Reformatting will also likely break the NNPDF pheno scripts in the paper repo.

pineappl_cli/src/plot.py
```diff
@@ -1,18 +1,39 @@
...
+ylog = False
+scales = 1
+enable_int = False
+enable_abs = False
+enable_rel_ewonoff = False
+enable_abs_pdfs = False
+enable_ratio_pdf = False
+enable_double_ratio_pdf = False
+enable_rel_pdfunc = False
+enable_rel_pdfpull = False
```
cschwan, line 24: Put these booleans together, maybe in a dictionary.

pineappl_cli/src/plot.py
```diff
@@ -71,24 +85,31 @@
...
-        {enable_abs}plot_abs,
-        {enable_rel_ewonoff}plot_rel_ewonoff,
-        {enable_abs_pdfs}plot_abs_pdfs,
-        {enable_ratio_pdf}plot_ratio_pdf,
-        {enable_double_ratio_pdf}plot_double_ratio_pdf,
-        {enable_rel_pdfunc}plot_rel_pdfunc,
-        {enable_rel_pdfpull}plot_rel_pdfpull,
-    ]
+    panels = []
+    if enable_int:
```
cschwan, line 89: All of these `if`s should be put together in a loop.

pineappl_cli/src/plot.rs
```diff
@@ -99,32 +106,34 @@ fn map_format_channel(
...
         .map(|(values, label)| {
             format!(
-                "                (
+                "            (
                     \"{}\",
                     np.array([{}]),
                     np.array([{}]),
                     np.array([{}]),
                 ),",
-                label.replace('_', r"\_"),
```
cschwan, line 127: This will lead to problems once you use an NNPDF sets; these contain underscores which  will break LaTeX.

cschwan (changes requested):

felixhekhorn:
> There are a bit too many changes doing different things in here, which make reviewing and fixing the tests difficult. Could you please focus first on one goal, namely making `plot.py` a proper Python file? We can reformat and lint later on when we know it's working. Reformatting will also likely break the NNPDF pheno scripts in the paper repo.

mmmmm ... [grummel] ... the commits so far are the minimal changes I would say, i.e. they make `plot.py` a Python file and the only thing on top, which is one of the main motivation of doing this thing, is `ruff` ... I don't know how one would do less

also I want to largely iterate the Python file to make it easier to maintain, to ban all the globals, and to make the plot functions reusable. Actually I already started on the process, but now I don't know whether to commit (and whether my last hour was wasted ...)

pineappl_cli/src/plot.py
```diff
@@ -1,18 +1,39 @@
...
+enable_int = False
+enable_abs = False
+enable_rel_ewonoff = False
+enable_abs_pdfs = False
+enable_ratio_pdf = False
+enable_double_ratio_pdf = False
+enable_rel_pdfunc = False
+enable_rel_pdfpull = False
+output = ""
+data = {}
```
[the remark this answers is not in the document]
felixhekhorn: and kept this! `data` and `metadata` are still at the bottom

pineappl_cli/src/plot.py
felixhekhorn: wait a second: you just told me I should rework so much! and in fact this is a prime example of what I want to rework

pineappl_cli/src/plot.py
felixhekhorn: [see above](https://github.com/NNPDF/pineappl/pull/312#discussion_r1768479509)

pineappl_cli/src/plot.rs
felixhekhorn: no, I don't think so. 1. this leads to
```
/home/felix/Physik/N3PDF/PineAPPL/pineappl/test-data/plot2.py:525: SyntaxWarning: invalid escape sequence '\_'
  "NNPDF31\_nlo\_as\_0118\_luxqed",
```
and 2. I didn't check the actual plot script yet (since it is too complicated), but this would only be true if the label would be "tex-ified", which I think it does not. I'm testing atm with simple case in the unit test and there the plot (-label) is just fine

felixhekhorn:
So, I committed nevertheless (because I don't want to loose my work).

I would say: if you want to have a smaller PR break this PR at [51a5d53](https://github.com/NNPDF/pineappl/pull/312/commits/51a5d536ae728535f9d2290613f4359b47647e4b). Now that may still be difficult too review, but this is because the previous state was bad - I'm 100% confident that this PR improves the situation. Yes, even that change can and may break the pheno paper plots, but then we can wait until that paper is out to merge - or maybe this PR reaches a state first, where whatever hacks were needed so far there are no longer needed. If you wish we can put the commits after the one mentioned into a separate PR - where I want to change lots of things.

Speaking of questionable content: what exactly is the purpose of `map_format_e_join_repeat_last`? if this is just for getting simpler arguments for matplotlib then I will ban that function and replicate the behaviour on the actual spot (if that is needed - because I believe sometimes it is counteracted again on the Python side)

Radonirinaunimi:
I agree that the previous version of the script was in a lousy state and I think that we should fix this once and for all here. Hence I believe it would be reasonable to expect a few changes. Skimming through I think @felixhekhorn's changes are going in the right direction and it seems that to make the tests pass it will suffice to "copy" the contents of `plot.py` into `tests/plot.rs`.

> may break the pheno paper plots

AFAIA, the pheno paper is not relying on this at all.

PS: I will have to wait for this to resume https://github.com/NNPDF/pineappl/pull/310.

alecandido:
A very stupid and almost irrelevant improvement:
image
```py
ax.set_ylim(0.0)
```
(just to avoid the bottom gap)

felixhekhorn:
> A very stupid and almost irrelevant improvement:

@alecandido Maybe this comment belongs to #309?

alecandido:
> @alecandido Maybe this comment belongs to #309?

Ah, yes, I confused two open tabs, sorry 😅

pineappl_cli/src/plot.py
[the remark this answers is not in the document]
cschwan: I just tested this and yes, that's indeed the case!

pineappl_cli/src/plot.py
cschwan: Done in commit 24b97b8cb46fdf5511a94219e7e40e53f74217af.

pineappl_cli/src/plot.rs
cschwan: Fixed in commit 68e73907cd8b30148ce3ab00bbee9087b1101971.

pineappl_cli/src/plot.py
felixhekhorn: Improved in 143b5d53 - `globals()` should be avoided if possible ...

cschwan: I believe `globals()` is OK since execute a generated script, but if we want to split parts of it into a library this is better.

cschwan:
Is there anything missing here? If not, I'll fix the tests and merge it into master.

Radonirinaunimi:
> Is there anything missing here? If not, I'll fix the tests and merge it into master.

Not as far as I know. Things LGTM.

Radonirinaunimi:
@felixhekhorn  Can you confirm?

felixhekhorn:
I still want to [copy the file to pineappl_py](https://github.com/NNPDF/pineappl/pull/312#issuecomment-2363034277), but this can also be done later, if you want to merge this quickly (to not block other PRs) (as I'm not sure how quickly I can come back to this).

Then at a later stage, I'd still like to improve the script as a whole and actually now that I'm thinking about it - some part of that will need to take place together with the move, i.e. it is not trivial, so let's postpone ...

Radonirinaunimi:
@cschwan In case you've not worked on this yet I've fixed the tests locally that I could push (if it helps).

cschwan:
Yes, please do!

Radonirinaunimi (approved):

cschwan:
Are there any immediate concerns that should be addressed? Otherwise I'm going to merge this into `master`.

Radonirinaunimi:
> Are there any immediate concerns that should be addressed? Otherwise I'm going to merge this into `master`.

This actually could be merged.

PR #324, opened 2025-01-27
Add alternative development using Nix

Radonirinaunimi:
Adds a documentation on alternatively building `pineappl` with `nix-shell`.


alecandido:
If you wish, there is also a flake ready (but not updated since a while) in #245.

You can use it following the instructions in https://github.com/NNPDF/pineappl/pull/245#issuecomment-1962892499, which mainly boil down to:
```sh
nix develop . --impure
```

Issue #327, opened 2025-01-31
Update Python tutorial for PineAPPL 1.0


t7phy:
See #299.

Issue #326, opened 2025-01-31
Polish code for PineAPPL 1.0


t7phy:
- [ ] Fill in all the TODO comments
- [ ] Cleanup code
- [ ] Raise code coverage

felixhekhorn:
this line
https://github.com/NNPDF/pineappl/blob/6c9e66727564ab80f7b138bc27b3d8aac1dad130/docs/cli-reference.md?plain=1#L68
surely is outdated - this may or may not be true also for other pages of the `docs` folder

PR #328, opened 2025-02-03
Stabilize APIs

Radonirinaunimi:
Addresses the API parts of https://github.com/NNPDF/pineappl/issues/326 and https://github.com/NNPDF/pineappl/issues/327 

- [x] Python part:
  - [x] Fix the tutorials 
  - [x] stabilize the API 
  - [x] increase test coverage
- [x] Fortran part (done modulo some potential memory leaks in examples):
  - [x] stabilize the API (add still missing functions)
  - [x] set examples `_v1` as default/main and deprecate the old ones 
  - [x] run Fortran examples on the CI
  - [x] fix bugs in the generalized convolutions
- [x] C/C++ part: 
  - [x] fix a bug in `grid_convolve`
  - [x] test convolutions with two different QED PDF sets
- [x] C++ OOP:
  - [x] implement generalized convolution function
  - [x] run C++ OOP examples on the CI


Radonirinaunimi:
@cschwan In case you want to have a look, this is complete.

felixhekhorn:
Could you please also adjust https://github.com/NNPDF/pineappl/blob/stabilize-apis/pineappl_capi/include/PineAPPL.hpp ?

cschwan:
Looks quite good to me, however we can still increase the CAPI code coverage:

- fill a `Grid` with no reweighting to cover [line 124](https://app.codecov.io/gh/NNPDF/pineappl/blob/stabilize-apis/pineappl_capi%2Fsrc%2Flib.rs#L124)
- call `pineappl_grid_key_value` with `initial_state_1` and `initial_state_2` to cover [lines 978 to 981](https://app.codecov.io/gh/NNPDF/pineappl/blob/stabilize-apis/pineappl_capi%2Fsrc%2Flib.rs#L979) - shouldn't this happen already?
- call `pineappl_grid_convolve` with `mu_scales` set to `nullptr` to cover [line 1836](https://app.codecov.io/gh/NNPDF/pineappl/blob/stabilize-apis/pineappl_capi%2Fsrc%2Flib.rs#L1836)

In the Python API we should

- trigger the error paths in [`Grid.set_bwfl`](https://app.codecov.io/gh/NNPDF/pineappl/blob/stabilize-apis/pineappl_py%2Fsrc%2Fgrid.rs#L225) and [`Grid.merge`](https://app.codecov.io/gh/NNPDF/pineappl/blob/stabilize-apis/pineappl_py%2Fsrc%2Fgrid.rs#L665)
- test the `boc` module more extensively.

cschwan:
@Radonirinaunimi BTW: the Fortran module file is shipped with PineAPPL and isn't installed. This means we can forgo giving any stability guarantees and just change it as we see fit. If a user wants to use an older version of the module they always can, because the module is just a thin wrapper of the CAPI which is stable. So if you think it's too much of an effort to keep the deprecated functions in the Fortran module you can just strip them out. The same holds in principle for the C++ OOP API that @felixhekhorn mentioned.

Radonirinaunimi:
> @Radonirinaunimi BTW: the Fortran module file is shipped with PineAPPL and isn't installed. This means we can forgo giving any stability guarantees and just change it as we see fit. If a user wants to use an older version of the module they always can, because the module is just a thin wrapper of the CAPI which is stable. So if you think it's too much of an effort to keep the deprecated functions in the Fortran module you can just strip them out.
> 
I've actually always wondered why we don't install the Fortran API? We can promote it to a folder `pineappl_fapi` and add versions to it? Are there reasons against this?

> The same holds in principle for the C++ OOP API that @felixhekhorn mentioned. I'd even be in favour of either completely removing the C++ OOP API or letting someone else maintain it, possibly in a different repository even, because in C++ you can always use the CAPI without any additional burden. The C++ OOP vs. C functions is really just a stylistic choice. But that's a decision you have to make, @Radonirinaunimi; having 4 different APIs (CAPI, Python, Fortran, C++ OOP) may be too much a maintainance burden.

As for this, I don't actually mind maintaining these APIs. The current situation is a bit peculiar because `v1` added lots of changes, but by now I think the main crate should be much more stable and future changes will not be as significant.

So, yes, I can also update this:
> Could you please also adjust https://github.com/NNPDF/pineappl/blob/stabilize-apis/pineappl_capi/include/PineAPPL.hpp?

after all of the following is done: https://github.com/NNPDF/pineappl/pull/328#issuecomment-2633846854.

Radonirinaunimi:
> test the boc module more extensively.

@cschwan, by this do you mean increasing the areas/percentage covered? Because `boc` is already extensively tested in [test_boc.py](https://github.com/NNPDF/pineappl/blob/stabilize-apis/pineappl_py/tests/test_boc.py). The problem is that I am not sure it is possible (from Python side) to hit the `enum` such as in [line 311](https://app.codecov.io/gh/NNPDF/pineappl/blob/stabilize-apis/pineappl_py%2Fsrc%2Fboc.rs#L311).

cschwan:
> I've actually always wondered why we don't install the Fortran API? We can promote it to a folder `pineappl_fapi` and add versions to it? Are there reasons against this?

There are a few reasons:
1. compiled Fortran module files depend not only on the compiler, but also on the compiler version
2. there's no agreed-upon installation directory where these files would go, see for instance the discussion here: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=49138>
3. even if you solve the first two problems, there's no (?) tooling to compile and install the modules, and we should even try to write our tool, which is bound to break for different compilers

cschwan:
> > test the boc module more extensively.
> 
> @cschwan, by this do you mean increasing the areas/percentage covered? Because `boc` is already extensively tested in [test_boc.py](https://github.com/NNPDF/pineappl/blob/stabilize-apis/pineappl_py/tests/test_boc.py). The problem is that I am not sure it is possible (from Python side) to hit the `enum` such as in [line 311](https://app.codecov.io/gh/NNPDF/pineappl/blob/stabilize-apis/pineappl_py%2Fsrc%2Fboc.rs#L311).

We should strive to get as close as possible to 100% coverage. Your last two commits did raise the coverage and I believe that last point you should check is whether `PyScales::default` is superfluous (I think it is).

pineappl_capi/include/PineAPPL.hpp
```diff
@@ -136,21 +196,119 @@ struct Order {
...
+  /**
+   * @brief Optimizes the grid representation for space efficiency.
+   */
+  void optimize() const { pineappl_grid_optimize(this->raw); }
+};
+
+/** @brief The central grid object. */
+struct Grid : public GridBase {
+  // /** @brief Underlying raw object. */
+  // pineappl_grid *raw;
```
felixhekhorn, line 310:
```suggestion
```

pineappl_capi/include/PineAPPL.hpp
```diff
@@ -119,15 +115,79 @@ struct Lumi {
...
+  /** @brief Number of elements. */
+  std::size_t count() const { return pineappl_channels_count(this->raw); }
+
+  /**
+   * @brief Add a channel function.
+   * @param c channel function
+   */
+  void add(const ChannelsEntry &c) const {
+    const std::size_t combinations = c.channels_entry.size();
+    const std::size_t nb_convolutions =
```
felixhekhorn, line 164:
```suggestion
    if (0 == combinations)
      return;
    const std::size_t nb_convolutions =
```
and we need more similar things - since in C++ people usually do sanity checks (unlike Python). E.g. make sure `c.channels_entry[0].entry[0]` exists, and all entries have the same size (can e.g. be done during the loop below)

pineappl_capi/include/PineAPPL.hpp
```diff
@@ -119,15 +115,79 @@ struct Lumi {
...
    * @param entry position in lumi
    */
   std::size_t combinations(const std::size_t entry) const {
     return pineappl_lumi_combinations(this->raw, entry);
   }
 };
 
+/** @brief Entry in sub-channel function.  */
+struct SubChannelEntry {
+  /** @brief First parton id. */
```
felixhekhorn, line 129: 1. since this is actually complicated we need some documentation here
2. this is a case of "inheritance vs. composition", not sure which one is better ... however, are there any additional methods or attributes in the first place? then also a simple `typedef` might be sufficient

pineappl_capi/include/PineAPPL.hpp
```diff
@@ -136,21 +196,119 @@ struct Order {
...
 
   /** @brief Exponent of the logarithm of the scale factor of the factorization
    * scale. */
   std::uint32_t logxif;
 };
 
-/** @brief The central grid object. */
-struct Grid {
+/** @brief Extension of `Order` to accommodate for v1 representation. */
+struct OrderV1 {
```
felixhekhorn, line 209: I would move all the `V1` objects in a `V1` namespace instead

pineappl_capi/include/PineAPPL.hpp
```diff
@@ -174,19 +333,6 @@ struct Grid {
...
 
-  Grid() = delete;
-
-  Grid(const Grid &) = delete;
-
-  Grid(Grid &&) = delete;
-
-  Grid &operator=(const Grid &) = delete;
-
-  Grid &operator=(Grid &&) = delete;
```
felixhekhorn, line 185: don't we need all those still? i.e. we want to deny copying any `Grid` object for users (and deny `GridBase` for developers)

Radonirinaunimi:
> > I've actually always wondered why we don't install the Fortran API? We can promote it to a folder `pineappl_fapi` and add versions to it? Are there reasons against this?
> 
> There are a few reasons:
> 
>     1. compiled Fortran module files depend not only on the compiler, but also on the compiler version
> 
>     2. there's no agreed-upon installation directory where these files would go, see for instance the discussion here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=49138
> 
>     3. even if you solve the first two problems, there's no (?) tooling to compile and install the modules, and we should even try to write our tool, which is bound to break for different compilers

I will come back to this later, but it seems now that there is a more pressing issue. I've found a bug in the C bindings that compute `pineappl_grid_convolve_with_two` and `pineappl_grid_convolve`.  In both, the `state` was always fixed and therefore always the same PDF is used.

@cschwan , given that all of the grids in our examples only contain photon-photon channel, would it be possible to add another QED set to this list:
https://github.com/NNPDF/pineappl/blob/dbae06ef071ca6a99f15be16b05c7ac0fd9d4581/maintainer/pineappl-ci/script.sh#L15-L23
in order to test convolutions with "different" initial states?

examples/object-oriented-cpp/dyaa.cpp
```diff
@@ -2,149 +2,176 @@
...
+  PidBasis pid_basis = Evol;
+  std::vector<std::int32_t> pids = {2212, 2212};
+
+  // Define the types of convolutions
+  ConvType h1 = UnpolPDF;
+  ConvType h2 = UnpolPDF;
+  std::vector<ConvType> convolution_types = {h1, h2};
+
+  // Define the Kinematics
+  Kinematics scales = {Scale, 0};
```
felixhekhorn, line 134: the type `Kinematics` is coming from the C-API, right? mmm ... I don't like that name floating around globally (because it might well be used [by others](https://github.com/felixhekhorn/LeProHQ/blob/5f0c8664db9b0c9cb3d483d62b68dd1ecba6526b/LeProHQ%2B%2B/include/FullyDiff/KinematicVars.hpp#L15)) ... can't we do something against? C has no namespaces so there is nothing we can do directly there, right? should we add the `#include <pineappl_capi>` _inside_ the `namespace PineAPPL`? a 2min google search says this is in general a bad idea - but I think for our case this is justified: because the two things are actually one and the same and there are bound together forever. If you break C you break also C++

felixhekhorn (commented):
as already said: if the OOP is too much effort please tell me and we can strip it from here

cschwan:
> [..] I've found a bug in the C bindings that compute `pineappl_grid_convolve_with_two` and `pineappl_grid_convolve`. In both, the `state` was always fixed and therefore always the same PDF is used.

We can change `state` to be an array of `void*` that must match the length of `xfxs` for `pineappl_grid_convolve`. Since we can't change `pineappl_grid_convolve_with_two` the user has to take care of this themselves; this can be done with code like the following:

```cpp
struct State {
    PDF pdfs[2];
};

double xfx1(int32_t x, double x, double q2, void* state) {
    auto* pdf = dynamic_cast <State*> (state)->pdfs[0];
    // ..
}

double xfx2(int32_t x, double x, double q2, void* state) {
    auto* pdf = dynamic_cast <State*> (state)->pdfs[1];
    // ..
}

int main() {
    State state;
    // ..
    pineappl_grid_convolve_with_two(..., xfx1, ..., xfx2, ..., &state);
    // ..
}
```
Was this what you were worried about? It's not too complicated, but I agree that for the new function we better increase the ergonomics.

> 
> @cschwan , given that all of the grids in our examples only contain photon-photon channel, would it be possible to add another QED set to this list:
> 
> https://github.com/NNPDF/pineappl/blob/dbae06ef071ca6a99f15be16b05c7ac0fd9d4581/maintainer/pineappl-ci/script.sh#L15-L23
> 
> in order to test convolutions with "different" initial states?

Sure, go ahead with this. After patching the script you can run the container action in this branch.

examples/object-oriented-cpp/dyaa.cpp
cschwan: You can't use `namespace` in C because that's not a keyword in C. It will also probably break the entire thing, because if you rename symbols the linker won't find them anymore. However, I agree that this is a problem that must be solved for the all new enums. I suggest we use `[export.rename]` in `cbindgen.toml` to take of this:
```toml
"ConvType" = "pineappl_conv_type"
"InterpMeth" = "pineappl_interp_meth"
```
and so on and so forth, so basically converting CamelCase to snake_case with a `pineappl_` prefix (that's basically namespaces in C).

pineappl_capi/include/PineAPPL.hpp
cschwan: Or we simply nuke everything `v0`. This can still be used by using an old header.

pineappl_capi/include/PineAPPL.hpp
Radonirinaunimi: Yes, that'd be the plan indeed. And these are the easiest parts of the whole process 😇

pineappl_capi/include/PineAPPL.hpp
Radonirinaunimi: > Or we simply nuke everything `v0`. This can still be used by using an old header.

I like this 😄 and I'd do this actually. In this way, we won't need a complex data structure with Grid inheritance.

pineappl_capi/include/PineAPPL.hpp
Radonirinaunimi: I guess this will be deprecated by https://github.com/NNPDF/pineappl/pull/328#discussion_r1944992515.

examples/object-oriented-cpp/dyaa.cpp
Radonirinaunimi: I fully agree that the best way is to have a mapping in `cbindgen.toml`.

Radonirinaunimi:
> > [..] I've found a bug in the C bindings that compute `pineappl_grid_convolve_with_two` and `pineappl_grid_convolve`. In both, the `state` was always fixed and therefore always the same PDF is used.
> 
> We can change `state` to be an array of `void*` that must match the length of `xfxs` for `pineappl_grid_convolve`. Since we can't change `pineappl_grid_convolve_with_two` the user has to take care of this themselves; this can be done with code like the following:
> 
> ```c++
> struct State {
>     PDF pdfs[2];
> };
> 
> double xfx1(int32_t x, double x, double q2, void* state) {
>     auto* pdf = dynamic_cast <State*> (state)->pdfs[0];
>     // ..
> }
> 
> double xfx2(int32_t x, double x, double q2, void* state) {
>     auto* pdf = dynamic_cast <State*> (state)->pdfs[1];
>     // ..
> }
> 
> int main() {
>     State state;
>     // ..
>     pineappl_grid_convolve_with_two(..., xfx1, ..., xfx2, ..., &state);
>     // ..
> }
> ```
> 
> Was this what you were worried about? It's not too complicated, but I agree that for the new function we better increase the ergonomics.
> 
Yes, the fix was rather easy (its Fortran interface, not so much). 

The only minor difference w.r.t. your suggestion was that I only kept one "true" Lambda `xfx` and instead passed `pdf1_state` and `pdf2_state`. If I had set `state` to be an array for `pineappl_grid_convolve_with_two` then that function would be the same as `pineappl_grid_convolve`.

On the Fortran node, @janw20, would you perhaps be available to help? There are only two functions to fix, `pineappl_grid_convolve_with_two` and `pineappl_grid_convolve` but they seem a bit tricky.

examples/object-oriented-cpp/dyaa.cpp
Radonirinaunimi: Done in [8db10ab](https://github.com/NNPDF/pineappl/pull/328/commits/8db10abd10f3bd9a74aaddd79c2d7e81181f3ac7).

pineappl_capi/src/lib.rs
```diff
@@ -501,29 +503,30 @@ pub unsafe extern "C" fn pineappl_grid_convolve_with_one(
...
     pdg_id1: i32,
-    xfx1: extern "C" fn(pdg_id: i32, x: f64, q2: f64, state: *mut c_void) -> f64,
     pdg_id2: i32,
-    xfx2: extern "C" fn(pdg_id: i32, x: f64, q2: f64, state: *mut c_void) -> f64,
+    xfx: extern "C" fn(pdg_id: i32, x: f64, q2: f64, state: *mut c_void) -> f64,
     alphas: extern "C" fn(q2: f64, state: *mut c_void) -> f64,
-    state: *mut c_void,
+    pdf1_state: *mut c_void,
+    pdf2_state: *mut c_void,
+    alphas_states: *mut c_void,
```
felixhekhorn, line 519:
```suggestion
    alphas_state: *mut c_void,
```
why is there a plural? this seems wrong - `alphas` is bound to a fixed object

felixhekhorn: actually in `pineappl_grid_convolve` it is a singular - so I assume it is just a typo here

cschwan:
@Radonirinaunimi: you can't change `pineappl_grid_convolve_with_two`, because this method has been part of the CAPI since 0.8.0. Otherwise you would break existing programs. Remember: once we've added a symbol (function, struct, typedef, ...) in the CAPI that has an official release, we can never change or remove that symbol.

cschwan:
This is a recommendation, but one to make your life easier: since you've removed the v0 parts of the C++ API you can also remove these parts from the Fortran API; if someone needs them, they just can use an older module file from the git.

pineappl_capi/src/lib.rs
Radonirinaunimi: > why is there a plural? this seems wrong - `alphas` is bound to a fixed object

Good catch! That definitely has to be singular.

> actually in pineappl_grid_convolve it is a singular - so I assume it is just a typo here

As for `pineappl_grid_convolve`, the `convolve` part is a verb actually so that should be fine.

Radonirinaunimi:
> @Radonirinaunimi: you can't change `pineappl_grid_convolve_with_two` (nor `pineappl_grid_convolute_with_two`), because this function has been part of the CAPI since 0.8.0. Therefore you'd break existing programs. Remember: once we've added a symbol (function, struct, typedef, ...) in the CAPI that has an official release, we can never change or remove that symbol.

Yes, that's right. However, this was definitely a bug in the sense that the old functions were computed something completely wrong, and so I though it'd be fine to break the symbol.

So considering the above and the following:
> This is a recommendation, but one to make your life easier: since you've removed the v0 parts of the C++ API you can also remove these parts from the Fortran API; if someone needs them, they just can use an older module file from the git.

I believed such a price would be fine (?).

pineappl_capi/include/PineAPPL.hpp
Radonirinaunimi: > this is a case of "inheritance vs. composition", not sure which one is better ... however, are there any additional methods or attributes in the first place? then also a simple typedef might be sufficient

Yes, `SubChannelEntry` has/can have additional methods/attributes (such as returning the combination of channels for a given entry) which one might want to expose in the future, so I'd say a generalized struct is better in this case.

Btw, the "main" interfaces should now be complete if you want to have a look/improve.

Radonirinaunimi:
I believe this is now "ready". It seems that there are also some memory leaks in the Fortran examples that I am not able to really track. Apart from this, everything else should be good.

cschwan (changes requested):

examples/cpp/advanced-convolution-deprecated.cpp
```diff
@@ -34,11 +34,17 @@ double alphas(double q2, void* state) {
...
     std::string filename = "drell-yan-rap-ll-deprecated.pineappl.lz4";
-    std::string pdfset = "NNPDF31_nlo_as_0118_luxqed";
+    std::string pdfset1 = "NNPDF31_nlo_as_0118_luxqed";
+    std::string pdfset2 = "MSHT20qed_nnlo";
 
     switch (argc) {
+    case 4:
+        pdfset1 = argv[2];
+        pdfset2 = argv[3];
+        // fall through
```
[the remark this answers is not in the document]
Radonirinaunimi: Very good good catch (I missed updating that)! It is now fixed in [b5e98c3](https://github.com/NNPDF/pineappl/pull/328/commits/b5e98c3486273d292e1abca066b4ddf12215a053).

PR #329, opened 2025-02-18
Check three convolutions involving time-like hadron

Radonirinaunimi:
Checks the case in which one has three convolutions that involve a Fragmentation Function in the CLI and in the Python API.


cschwan:
Instead of testing a convolution, I suggest we test the scale variations. This is more general and should cover more lines.

Radonirinaunimi:
> Instead of testing a convolution, I suggest we test the scale variations. This is more general and should cover more lines.

Do you mean with proper scale-varied predictions/grids? So far, https://github.com/NNPDF/sihp-pp cannot track the `lr`, `lf`, and `la` orders and there isn't a plan to do so soon. So if we want to test this, I'd say we can do so with maybe the SIA and SIDIS grids.

cschwan:
I meant what I did in commit c4f85b939e64ce931880919fc0c477cabdfde5a4, but given that there's no log-xia grid you're right that this isn't very useful.

Radonirinaunimi:
So, could this be merged?

Issue #330, opened 2025-02-24
Allow CLI `evolve` to take a generic number of EKOs


Radonirinaunimi:
Currently,`pineappl evolve` still only takes up to two EKOs in which the second one is passed through the flag `--ekob`. This should be generalized to take a generic number of EKOs.

cc https://github.com/NNPDF/pineko/pull/215

cschwan:
Yes, that was an oversight, thanks for catching it! Due to limitations of `clap` we should provide a comma-separated list, which looks a bit odd for file names, but I believe that's our only option.

Radonirinaunimi:
> Due to limitations of `clap` we should provide a comma-separated list, which looks a bit odd for file names, but I believe that's our only option.

Yes, this was also my thinking. I can take care of implementing it.

PR #331, opened 2025-02-25
Allow a generic number of `EKOs` to be passed to `pineappl evolve`

Radonirinaunimi:
This addresses #330.

Things to do:
- [x] rename `EkoNames` to `EkoPaths`
- [x] ~find a robust and stable way to parse multiple EKOs~
- [x] rely on information inside the EKOs to derive the types of convolutions
- [x] add minor tutorial on the use of `p,f,pf,fp`markings in multi-convolution

### DIFF
--- a/docs/cli-tutorial.md
+++ b/docs/cli-tutorial.md
@@ -60,6 +60,44 @@ Let's have a closer look at what the output shows:
 - the last column `dsig/detal` shows the differential cross section (`dsig`)
   for the bin limits given in the previous two columns.
 
+In the case where the grid requires different types of convolution functions,
+the LHAPDF names need to be appended with `+p`, `+f`, or `+pf` (equivalently
+`+fp`) which respectively correspond to polarized, fragmentation function, or
+both.
+
+As an example, let's download a grid that contains one polarized and one
+unpolarized proton:
+```
+wget 'https://data.nnpdf.science/pineappl/test-data/STAR_WMWP_510GEV_WM-AL-POL.pineappl.lz4'
+```
+
+Then, to perform the convolution, run the following command:
+```
+pineappl convolve STAR_WMWP_510GEV_WM-AL-POL.pineappl.lz4 NNPDFpol11_100+p,CT18NNLO
+```
+
+where we specified the PDF set `NNPDFpol10_100` to be polarized. Note that
+the order in which the PDF sets are passed do not matter; `pineappl` will
+automatically identify which object should be convolved with a given PDF set.
+
+If successful, the command above should print the following result (again the
+numbers don't really matter):
+```
+b                  eta                   dsig/deta
+                   []                      [pb]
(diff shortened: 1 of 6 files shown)


Radonirinaunimi:
> I'd make this a bit simpler by relying more on clap: replace `EkoPaths` with `Vec<EkoPath>` and make `EkoPath` a struct of a `String` and a `ConvType`. Furthermore, use `#[arg(value_delimiter = ',', num_args = 1)]`. Then clap does most of the parsing.

So you want to remove `EKOs` as a positional argument? Because I don't think it would work otherwise.

cschwan:
No, it should remain a positional parameter. It's exactly one argument, but internally clap fills the `Vec` by splitting the argument by `,`.

cschwan:
On second thought, the information in the `+p`, `+f` and `+pf` syntax should be encoded in EKO's yaml files. We should rely on them instead of the information provided by the user.

Radonirinaunimi:
> No, it should remain a positional parameter. It's exactly one argument, but internally clap fills the `Vec` by splitting the argument by `,`.

But this, I think, doesn't work. `#[arg(value_delimiter = ',')]` is functionally equivalent to `.num_args(1..)` so it can only be the last positional argument and no positional arguments can come after.

The alternative would be to use named optional argument `#[arg(short, long, value_delimiter = ',')]`, but I am not sure we also want this.

> On second thought, the information in the +p, +f and +pf syntax should be encoded in EKO's yaml files. We should rely on them instead of the information provided by the user.

Yes, I will double check this to be certain, but if that's the case then that should definitely be the way to do it (added in the TODO of the description).

felixhekhorn:
> On second thought, the information in the `+p`, `+f` and `+pf` syntax should be encoded in EKO's yaml files. We should rely on them instead of the information provided by the user.

they are of course

just to say: if only we had someone work on https://github.com/NNPDF/eko/issues/97

docs/cli-tutorial.md
```diff
@@ -60,6 +60,44 @@ Let's have a closer look at what the output shows:
 - the last column `dsig/detal` shows the differential cross section (`dsig`)
   for the bin limits given in the previous two columns.
 
+In the case where the grid requires different types of convolution functions,
+the LHAPDF names need to be appended with `+p`, `+f`, or `+pf` (equivalently
+`+fp`) which respectively correspond to polarized, fragmentation function, or
+both.
```
felixhekhorn, line 66:
```suggestion
`+fp`) which respectively correspond to a polarized PDF, a fragmentation function, or
a polarized fragmentation function.
```

cschwan:
> But this, I think, doesn't work. `#[arg(value_delimiter = ',')]` is functionally equivalent to `.num_args(1..)` so it can only be the last positional argument and no positional arguments can come after.

Yes, I didn't realize this sooner, had to see it for myself. Thank for trying it out!

Issue #334, opened 2025-03-13
Problem inferring the convolution types of empty grids in old format


Radonirinaunimi:
See also discussion around https://github.com/NNPDF/pineko/pull/206#issuecomment-2721388644.

With numerical FONLL, intermediate positivity grids computed with `FFNS{3,6}-massive` are completely empty. This runs into issues if the grids were generated with a PineAPPL version before `0.8.6` because the types of convolutions are inferred from `grid.channels()`:

https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/pineappl/src/v0.rs#L294-L304

This then means that the returned values are always `None` and as a result corrupt the construction of the interpolation object:

https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/pineappl/src/v0.rs#L142-L145

The assertion below is therefore not satisfied and raises an error:

https://github.com/NNPDF/pineappl/blob/279ea0fe7a6636b99bcc00c4c611758634e94259/pineappl/src/grid.rs#L113-L119

As example of such a grid is available [here](https://data.nnpdf.science/pineko/theory_productions/data/grids/4000800503/NNPDF_POS_F2D_40.pineappl.lz4).

felixhekhorn:
I'd say empty grids are perfectly acceptable grids - they are just the 0 element in the algebra. And given the fact that observables may be decomposed into several grids (either for physics reason á là FONLL or just or numerical reasons), we need to support this case.

Given the fact this is clearly an edge case (which we still have to support), why don't we just add an exception to the logic?
e.g. make a marger `is_old: bool = false`, set that `true` in the mentioned switch, and then finally just do a `if (is_old && grid.is_empty()) { return default; }` where `default = vec![Some(Conv::new(ConvType::UnpolPDF, 2212)); 2]`

Radonirinaunimi:
> Given the fact this is clearly an edge case (which we still have to support), why don't we just add an exception to the logic? e.g. make a marger `is_old: bool = false`, set that `true` in the mentioned switch, and then finally just do a `if (is_old && grid.is_empty()) { return default; }` where `default = vec![Some(Conv::new(ConvType::UnpolPDF, 2212)); 2]`

Yes, I am thinking more or less the same. I think it is ok to assume that if a grid is:
- old - which they are if they pass through those codes
- empty
- and `initial_state_{index}: 2212`

[...]

PR #335, opened 2025-03-14
Infer the convolution types in empty grids with old format

Radonirinaunimi:
A simple fix to address https://github.com/NNPDF/pineappl/issues/334.

### DIFF
--- a/pineappl/src/v0.rs
+++ b/pineappl/src/v0.rs
@@ -291,17 +291,22 @@ fn read_convolutions_from_metadata(grid: &GridV0) -> Vec<Option<Conv>> {
                                 .map(|s| s.parse::<i32>())
                             {
                                 Some(Ok(pid)) => {
-                                    let condition = !grid.channels().iter().all(|entry| {
-                                        entry.entry().iter().all(|&(a, b, _)|
-                                            match index {
-                                                1 => a,
-                                                2 => b,
-                                                _ => unreachable!()
-                                            } == pid
-                                        )
-                                    });
+                                    // Addresses: https://github.com/NNPDF/pineappl/issues/334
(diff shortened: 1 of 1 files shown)


pineappl/src/v0.rs
```diff
@@ -291,17 +291,22 @@ fn read_convolutions_from_metadata(grid: &GridV0) -> Vec<Option<Conv>> {
...
-                                            match index {
-                                                1 => a,
-                                                2 => b,
-                                                _ => unreachable!()
-                                            } == pid
-                                        )
-                                    });
+                                    // Addresses: https://github.com/NNPDF/pineappl/issues/334
+                                    if grid.channels().is_empty() && pid == 2212 {
+                                        Some(Conv::new(ConvType::UnpolPDF, 2212))
```
cschwan, line 296: To answer my own question: we probably have this to catch initial-state electrons that don't correspond to convolutions.

cschwan (changes requested):
This looks good to me, but we need a regression test:
1. Upload one of the failing grids to the NNPDF server
2. Download the grid in the download script
3. Increase the version of the cache key
4. Write a test that performs a convolution with the empty grid

pineappl/src/v0.rs
Radonirinaunimi: Exactly! And this is actually the case for the empty positivity grid.

cschwan (changes requested):
After having implemented these tiny changes feel free to merge it!

Radonirinaunimi:
> After having implemented these tiny changes feel free to merge it!

Thanks a lot! Addressed all of them in [3860f5f](https://github.com/NNPDF/pineappl/pull/335/commits/3860f5fdc88521aea7ec3710ad9e1b1c29fe127f).

PR #336, opened 2025-03-14
Try to create Windows CLI wheels

cschwan:


cschwan:
While the build chain now works, the remaining problem is that the generated executable requires `libstdc++6.dll`, which we don't distribute.

PR #337, opened 2025-04-08
Exposes more functions to C-API for APFEL++

Radonirinaunimi:
Exposes more functions to the C-API for APFEL++, as discussed at DIS, in order to achieve two things:

- [x] Interface APFEL++ with PineAPPL in order to get (SI)DIS and SIA predictions (everything is already available)
- [x] Provide APFEL++ with an interface to read the contents of a PineAPPL grid and perform the evolution on the fly. Various ingredients are needed for this:
  - [x] functions to get and rotate the particle ID representation
  - [x] function that returns the list of channel combinations (see [display-channels.cpp](https://github.com/NNPDF/pineappl/blob/master/examples/cpp/display-channels.cpp))
  - [x] function that returns the contents for the subgrids for a given bin, order, and channel (see [get-subgrids.cpp](https://github.com/NNPDF/pineappl/blob/apfelxx-capi/examples/cpp/get-subgrids.cpp))

cc @vbertone, @enocera


pineappl_capi/src/lib.rs
```diff
@@ -895,6 +895,42 @@ pub unsafe extern "C" fn pineappl_grid_split_lumi(grid: *mut Grid) {
...
 }
 
+/// Change the particle ID representation of a given Grid.
+///
+/// # Safety
+///
+/// If `grid` does not point to a valid `Grid` object (for example when `grid` is the null pointer)
+/// or if `pid_basis` doe not refer to a correct basis, then this function is not safe to call.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_rotate_pid_basis(grid: *mut Grid, pid_basis: PidBasis) {
```
cschwan, line 905: I don't think it documented it anywhere, but the naming scheme for the CAPI is: `pineappl_{struct-name}_{method-name}`. So this function should be `pineappl_grid_rotate_pid_basis`.

pineappl_capi/src/lib.rs
```diff
@@ -895,6 +895,42 @@ pub unsafe extern "C" fn pineappl_grid_split_lumi(grid: *mut Grid) {
     grid.split_channels();
 }
 
+/// Change the particle ID representation of a given Grid.
```
cschwan, line 898: I'd use 'basis' instead of 'representation'.

pineappl_capi/src/lib.rs
```diff
@@ -895,6 +895,42 @@ pub unsafe extern "C" fn pineappl_grid_split_lumi(grid: *mut Grid) {
...
+/// If `grid` does not point to a valid `Grid` object, for example when `grid` is the `NULL`
+/// pointer, this function is not safe to call. The parameter `key` must be non-`NULL` and a valid
+/// C string.
+///
+/// # Panics
+///
+/// This function might panic if `pid_basis` contains interior null bytes. This occurs if  a new
+/// variant is added to the `PidBasis` enum in the future and the match statement isn't updated.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_get_pid_basis(grid: *mut Grid) -> *mut c_char {
```
cschwan, line 924: This function should be called `pineappl_grid_pid_basis`. Even more importantly, I'd change the return type to `PidBasis` to mirror the behaviour of `Grid::pid_basis`. The user can then use the result and `switch`/`case` to translate the enum-value to a string if that's needed.

pineappl_capi/src/lib.rs
```diff
@@ -895,6 +895,42 @@ pub unsafe extern "C" fn pineappl_grid_split_lumi(grid: *mut Grid) {
...
+/// If `grid` does not point to a valid `Grid` object (for example when `grid` is the null pointer)
+/// or if `pid_basis` doe not refer to a correct basis, then this function is not safe to call.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_rotate_pid_basis(grid: *mut Grid, pid_basis: PidBasis) {
+    let grid = unsafe { &mut *grid };
+
+    grid.rotate_pid_basis(pid_basis);
+}
+
+/// Get the particle ID representation of a Grid.
```
cschwan, line 911: Same comment as above.

cschwan (changes requested):

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
+    let mut subgrid = grid.subgrids()[[order, bin, channel]].clone();
+
+    let subgrid_shape = if subgrid.is_empty() {
+        let subgrid_dim = grid.convolutions().len() + 1;
+        &vec![0; subgrid_dim]
+    } else {
+        subgrid.shape()
+    };
+
+    let shape = unsafe { slice::from_raw_parts_mut(shape, grid.convolutions().len() + 1) };
```
cschwan, line 1937: That should be `grid.kinematics().len()` instead of `grid.convolutions().len() + 1`, because we may (and do) have grids with more than one scale.

Given that, we probably need to provide `pineappl_grid_kinematics_len`.

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
 }
+
+/// Get the number of convolutions for a given Grid.
+///
+/// # Safety
+///
+/// If `grid` does not point to a valid `Grid` object, for example when `grid` is the null pointer,
+/// this function is not safe to call.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_convolutions_len(grid: *mut Grid) -> usize {
```
cschwan, line 1906: This should be `pineappl_grid_convolutions_len`.

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
+/// Get the shape of a subgrid for a given bin, channel, and order.
+///
+/// # Safety
+///
+/// If `grid` does not point to a valid `Grid` object, for example when `grid` is the null pointer,
+/// this function is not safe to call. Additionally, the pointer that specifies the shape of the
+/// subgrid must be an array whose size must be `(n+1)` where `n` represents the number of
+/// convolutions as given by `pineappl_convolutions_len`.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_subgrid_shape(
```
cschwan, line 1920: This should be `pineappl_grid_subgrid_shape`.

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
+/// Get the subgrid for a given bin, channel, and order
+///
+/// # Safety
+///
+/// If `grid` does not point to a valid `Grid` object, for example when `grid` is the null pointer,
+/// this function is not safe to call. Additionally, the pointer that specifies the size of the subgrid
+/// when flattened must be an array; its size must be computed by multiplying the shape dimension as
+/// given by `pineappl_subgrid_shape`.
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_subgrid_array(
```
cschwan, line 1950: This should be `pineappl_grid_subgrid_array`.

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
+    grid: *mut Grid,
+    bin: usize,
+    order: usize,
+    channel: usize,
+    subgrid_array: *mut f64,
+) {
+    let grid = unsafe { &mut *grid };
+    let mut subgrid = grid.subgrids()[[order, bin, channel]].clone();
```
cschwan, line 1958: I don't think you need to clone the subgrid.

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
+
+    let flattened_shape = subgrid.shape().iter().copied().product();
+    let mut flattened_subgrid_array = vec![0.0; flattened_shape];
```
cschwan, line 1961: Remove this line and instead replace it with the line defining `subgrid_array` below. Then we save copying the entire array one time.

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
+#[no_mangle]
+pub unsafe extern "C" fn pineappl_subgrid_shape(
+    grid: *mut Grid,
+    bin: usize,
+    order: usize,
+    channel: usize,
+    shape: *mut usize,
+) {
+    let grid = unsafe { &mut *grid };
+    let mut subgrid = grid.subgrids()[[order, bin, channel]].clone();
```
cschwan, line 1928: Try to avoid the `.clone()` call.

pineappl_capi/src/lib.rs
```diff
@@ -1867,3 +1895,77 @@ pub unsafe extern "C" fn pineappl_grid_convolve(
...
+
+    let subgrid_shape = if subgrid.is_empty() {
+        let subgrid_dim = grid.convolutions().len() + 1;
```
cschwan, line 1931: This must be `grid.kinematics().len()`.

cschwan (changes requested):

pineappl_capi/src/lib.rs
Radonirinaunimi: Good point! You're completely right, that was an oversight. Fixed now in [bd68162](https://github.com/NNPDF/pineappl/pull/337/commits/bd68162dd7c88b9961ccbbf9641804cf2a0a2764)!

pineappl_capi/src/lib.rs
Radonirinaunimi: Still takes a while to fully get used to the naming convention 🙈, mainly because I was not fully consistent before in some places.

pineappl_capi/src/lib.rs
Radonirinaunimi: I think it is needed because `SubgridEnum` doesn't implement a `Copy` trait by default.

pineappl_capi/src/lib.rs
Radonirinaunimi: See [bd68162](https://github.com/NNPDF/pineappl/pull/337/commits/bd68162dd7c88b9961ccbbf9641804cf2a0a2764).

pineappl_capi/src/lib.rs
cschwan: I tried it myself and the problem was that `shape` was wrongly declared as `fn shape(&mut self)`. Since it's a getter `fn shape(&self)` should and does suffice, but that was the reason you really needed to clone. I fixed this in commit 3a9f23b3a22a213b4f614e05386909f65051e06b.

pineappl_capi/src/lib.rs
cschwan: Fixed with commit 3a9f23b3a22a213b4f614e05386909f65051e06b.

pineappl_capi/src/lib.rs
cschwan: I removed the superfluous indirection in commit a02beae383409569e78c4684d1939ae9af69883a.

cschwan:
@Radonirinaunimi can you please write another small test case, in which you call `pineappl_grid_subgrid_array` on an empty subgrid? The call is enough, and the test should make sure it doesn't panic/abort.

Radonirinaunimi:
Actually, just did it now in [7797940](https://github.com/NNPDF/pineappl/pull/337/commits/77979408bbb8bc5636973e51a70b705ee790cf3c), as it was just 2 lines.

examples/cpp/advanced-filling.cpp
```diff
@@ -6,6 +6,7 @@
 #include <iomanip>
 #include <memory>
 #include <vector>
+#include <numeric>
```
cschwan, line 9: What do we need `numeric` for?

Radonirinaunimi: Sorry, good catch! It was imported when I wanted to compute the shape of the subgrid as in here:
https://github.com/NNPDF/pineappl/blob/9c08d4f8c80f9ee797f6168adafef663b5ca70bc/examples/cpp/get-subgrids.cpp#L69-L70
which in the end I decided to hard-code but then forgot to remove.

cschwan (approved):

pineappl_capi/src/lib.rs
```diff
@@ -1745,12 +1773,13 @@ pub unsafe extern "C" fn pineappl_grid_split_channels(grid: *mut Grid) {
 pub unsafe extern "C" fn pineappl_channels_entry(
     channels: *const Channels,
     entry: usize,
+    nb_convolutions: usize,
```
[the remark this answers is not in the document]
cschwan: I think we can, see commit 9d6f89761f17805ccc47bc42a9358e5cd8b4c27b.

cschwan:
@Radonirinaunimi please have a look at commit 9d6f89761f17805ccc47bc42a9358e5cd8b4c27b. If you think that makes sense, feel free to merge!

Radonirinaunimi:
> @Radonirinaunimi please have a look at commit [9d6f897](https://github.com/NNPDF/pineappl/commit/9d6f89761f17805ccc47bc42a9358e5cd8b4c27b). If you think that makes sense, feel free to merge!

Yes, that indeed makes perfect sense, thanks! I will merge once the test are done.

PR #341, opened 2025-04-12
Polish codes

Radonirinaunimi:
Toward addressing #326 $-$ which would be good to have prior to releasing the stable `v1` version.


felixhekhorn:
maybe consider involving @ramonpeter ? since documentation is usually a good onboarding task

Radonirinaunimi:
> maybe consider involving @ramonpeter ? since documentation is usually a good on-boarding task

That is a very good idea!

ramonpeter:
Not quite sure what, but just contact me for more details

Radonirinaunimi:
@ramonpeter, the main idea would be to add documentation where TODO's appear (and optionally improve the codes where there are also TODO's).

PR #348, opened 2025-06-01
Materials for the PineAPPL v1 paper

Radonirinaunimi:
This PR adds minor modifications and new changes needed for the PineAPPL v1 paper. In particular, it adds:
-  [ ] two more scale variation prescription in presence of fragmentation scale:
    - [ ] **7-point prescription:** similar to the 7-point prescription in the case of two scale with the factorization and fragmentation scales correlated $\left( \mu_F = \mu_A \right)$. This is for example used in https://arxiv.org/pdf/2405.17540 and https://arxiv.org/pdf/2404.08597
    - [ ] **13-point prescription:** all the three scales are varied independently, spanning all the combinations except the extreme diagonal ones (where the ratio between scales $\mu_i / \mu_j = 4$ or $1/4$). This would be the analogous of the standard 7-point prescription in the case of three scales, and this is what we'd use for example to estimate MHOUs in PDF fits.
-  [ ] two examples (in C++ and Python) that illustrate the filling of LO predictions for a single-inclusive pion production in proton-proton collision. Here, we should check that the Monte Carlo simulations (esp. the kinematic parts) make sense.

### DIFF
--- a/examples/cpp/fill-grid-pp2hadron.cpp
+++ b/examples/cpp/fill-grid-pp2hadron.cpp
@@ -3,10 +3,8 @@
 #include <pineappl_capi.h>
 
 #include <cassert>
-#include <cmath>
 #include <cstddef>
 #include <random>
-#include <string>
 #include <vector>
 
 struct Psp2to2Hadron {
@@ -21,12 +19,11 @@ struct Psp2to2Hadron {
     double jacobian;
 };
 
-double me_gg2qqbar(double s, double t, double u) {
-    (void) s; // ignore dummy variable
+double me_gg2qqbar(double /*s*/, double t, double u) {
     double as2 = 0.118 * 0.118;
-    double PI2 = M_PI * M_PI;
+    double PI2 = std::acos(-1.0) * std::acos(-1.0); // π^2
     // TODO: double-check
-    return (16 * PI2 * as2 / 6.0) * (std::pow(u, 2) + std::pow(t, 2)) / (u * t);
+    return (16 * PI2 * as2 / 6.0) * (u * u + t * t) / (u * t);
 }
 
 Psp2to2Hadron pspgen_pp2hadron(std::mt19937& rng, double mmin,
@@ -186,7 +183,7 @@ int main() {
     // choose the Evolution Basis to represent the grid
     pineappl_pid_basis pid_basis = PINEAPPL_PID_BASIS_EVOL;
     // define the types of hadrons and set them to be Unpolarised
-    pineappl_conv convs[] = {
+    std::vector<pineappl_conv> convs = {
         {PINEAPPL_CONV_TYPE_UNPOL_PDF, 2212}, // proton
         {PINEAPPL_CONV_TYPE_UNPOL_PDF, 2212}, // proton
         {PINEAPPL_CONV_TYPE_UNPOL_FF, 211},   // pion
@@ -197,7 +194,7 @@ int main() {
     pineappl_kinematics x1 = {PINEAPPL_KINEMATICS_X, 0};
     pineappl_kinematics x2 = {PINEAPPL_KINEMATICS_X, 1};
     pineappl_kinematics z = {PINEAPPL_KINEMATICS_X, 2};
-    pineappl_kinematics kinematics[4] = {scales, x1, x2, z};
+    std::vector<pineappl_kinematics> kinematics = {scales, x1, x2, z};
 
     // define the specificities of the interpolations `interpolations = (μ, x1, x2, z)`
     pineappl_reweight_meth scales_reweight = PINEAPPL_REWEIGHT_METH_NO_REWEIGHT;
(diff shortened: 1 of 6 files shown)


examples/cpp/fill-grid-pp2hadron.cpp
```diff
@@ -0,0 +1,245 @@
...
+    double x1;
+    double x2;
+    double z;
+    double pt_hadron;
+    double y_hadron;
+    double jacobian;
+};
+
+double me_gg2qqbar(double s, double t, double u) {
+    (void) s; // ignore dummy variable
```
cschwan, line 25: Ignore this as described in <https://github.com/NNPDF/pineappl/pull/344#discussion_r2124438629>.

examples/cpp/fill-grid-pp2hadron.cpp
```diff
@@ -0,0 +1,245 @@
...
+    double as2 = 0.118 * 0.118;
+    double PI2 = M_PI * M_PI;
```
cschwan, line 27: Do not use `M_PI`, as it's not part of any C++ standard. You can use `std::acos(-1.0)`, which GCC evaluates at compile time.

examples/cpp/fill-grid-pp2hadron.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // TODO: double-check
+    return (16 * PI2 * as2 / 6.0) * (std::pow(u, 2) + std::pow(t, 2)) / (u * t);
```
cschwan, line 29: Don't use `std::pow(a, b)` for small integer arguments `b`; this will result in slow code as it's basically equivalent to `std::exp(std::log(a) * b)`. Use `u * u` and `t * t` instead, that's also shorter to write.

examples/cpp/fill-grid-pp2hadron.cpp
```diff
@@ -0,0 +1,245 @@
...
+        {PINEAPPL_CONV_TYPE_UNPOL_PDF, 2212}, // proton
+        {PINEAPPL_CONV_TYPE_UNPOL_FF, 211},   // pion
+    };
+
+    // define the kinematics object `kinematics = (μ, x1, x2, x)`
+    pineappl_kinematics scales = {PINEAPPL_KINEMATICS_SCALE, 0};
+    pineappl_kinematics x1 = {PINEAPPL_KINEMATICS_X, 0};
+    pineappl_kinematics x2 = {PINEAPPL_KINEMATICS_X, 1};
+    pineappl_kinematics z = {PINEAPPL_KINEMATICS_X, 2};
+    pineappl_kinematics kinematics[4] = {scales, x1, x2, z};
```
cschwan, line 200: If you initialize arrays like this, leave out the number of elements, which the compiler automatically infers. This is beneficial if you change it.

examples/cpp/fill-grid-pp2hadron.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // define the values of the unphysical scales `mu_scales = (μR, μF, μD)`
+    // where here we do not consider the fragmentation scale μD
+    pineappl_scale_func_form scale_mu = {PINEAPPL_SCALE_FUNC_FORM_SCALE, 0};
+    pineappl_scale_func_form mu_scales[3] = {scale_mu, scale_mu, scale_mu};
+
+    // ---
+    // Create the grid, fill it with Monte Carlo weights, and dump into disk
+
+    auto* grid = pineappl_grid_new2(bins.size() - 1, bins.data(), orders.size() / 5,
+        orders.data(), channels, pid_basis, convs, nb_convolutions + 1,
```
cschwan, line 224: Instead of `nb_convolutions + 1` you could write `sizeof (interpolations)`, which documents that this parameter just gives the length of the next array. This is also the safer choice, if someone decides to change the interpolation dimensions.

examples/cpp/fill-grid-pp2hadron.cpp
```diff
@@ -0,0 +1,245 @@
...
+    fill_grid(grid, 100000);
+
+    // add some metadata to the grid
+    pineappl_grid_set_key_value(grid, "x1_label", "pT");
+    pineappl_grid_set_key_value(grid, "y_label", "dsig/dpT");
+    pineappl_grid_set_key_value(grid, "x1_unit", "GeV");
+    pineappl_grid_set_key_value(grid, "y_unit", "pb/GeV");
+
+    // write the grid into disk
+    std::string filename = "pp2hadron-pt.pineappl.lz4";
```
cschwan, line 240: Inline the `char const*` below.

examples/cpp/fill-grid-pp2hadron.cpp
```diff
@@ -0,0 +1,245 @@
...
+        pids.insert(pids.end(), {21, 21, i});
+        factors.push_back(1.0);
+    }
+    pineappl_channels_add(channels, pids.size() / nb_convolutions, pids.data(),
+        factors.data());
+
+    // specify the perturbative orders that will be filled into the grid
+    // orders specifies the power of the tuple `orders = (αs, α, lR, lF, lD)`
+    // in this example, we only fill the LO QCD
+    std::vector<uint8_t> orders = {1, 0, 0, 0, 0};
```
cschwan, line 175: Below you use C-arrays for other parameters, so you should decide between one of them and `std::vector` IMO. Although there are also `std::array` and `std::unique_ptr<[]>` available. I think in this instance C-arrays are probably the best because you don't allocate.

cschwan (changes requested):

examples/cpp/fill-grid-pp2hadron.cpp
Radonirinaunimi: Yes, I knew. Still, it was the OCD part of me 🙈

examples/cpp/fill-grid-pp2hadron.cpp
Radonirinaunimi: In order to be consistent, and for simplicity because we most of the time need to know the actual size of the objects, I decided to go with `std::vector`.

Radonirinaunimi:
Closing because deprecated.

PR #350, opened 2025-06-12
Make `pineappl evolve` support EKO `v0.15`

Radonirinaunimi:
As the title says, `pineappl evolve` does not work with EKO v0.15 because of a change in format. Examples of a grid and an EKO (`LHCB_WP_8TEV.*`) have been uploaded in the usual place on the server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- added a new `V3` metadata reader to the `pineappl evolve` CLI for EKOs
+  generated with `v0.15.0` or higher
+- C API: added new functions `pineappl_grid_evolve_info_shape`,
+  `pineappl_grid_evolve_info`, and `pineappl_grid_evolve` to evolve grids
+- C API: added `pineappl_fktable_optimize` to optimize FK Table-like objects
+  given an optimization assumption
+
 ## [1.0.0] - 10/06/2025
 
 PineAPPL 1.0 is a major rewrite from the previous version, allowing grids to


cschwan:
Depends on what `metadata.xgrid` actually is. Is it x-grid for the input grids or for the output FK-tables? It must be the grid x-grid. Another possibility: have you re-interpolated the EKO? If you don't do that, x-grids probably don't match.

cschwan:
I always had to load the EKO in Python, reinterpolate it for the given x-grid, store it, and only that operator is usable with the CLI. I'll try find my Pyrhon script later this evening.

cschwan:
I pushed it to master, hopefully it still works. You may want to change the hardcoded paths, though.

Radonirinaunimi:
> I pushed it to master, hopefully it still works. You may want to change the hardcoded paths, though.

Thanks! Unfortunately, the script doesn't work anymore with new EKO versions. I opened an issue here: https://github.com/NNPDF/eko/issues/472

Once a fix is provided, I will fix the script in `master`. I can write it such it is compatible with all the versions of EKO using https://github.com/vmoens/pyvers.

cschwan:
> Once a fix is provided, I will fix the script in `master`. I can write it such it is compatible with all the versions of EKO using https://github.com/vmoens/pyvers.

Do we need that backwards compatibility? If not there's no need to make it more complicated.

pineappl_cli/src/evolve.rs
```diff
@@ -70,6 +70,7 @@ mod eko {
         V0(MetadataV0),
         V1(MetadataV1),
         V2(MetadataV2),
+        V3(MetadataV3),
```
cschwan, line 73: @felixhekhorn would probably say that we should explicitly encode the EKO version here, and I tend to agree. However, I didn't do that in the past, because I didn't know what EKO version ranges `V0`, `V1` and `V2` correspond to, so let's at least note that here by adding a comment:

    // v0.15 - v????

When we know all the ranges we change that in separate PR.

pineappl_cli/src/evolve.rs
```diff
@@ -325,6 +338,51 @@ mod eko {
...
+                    fac1.insert(file_stem, op_info.scale);
+                } else if path.as_os_str() == "./operator.yaml" {
+                    operator = Some(serde_yaml::from_reader(entry)?);
+                }
+            }
+
+            let operator =
+                operator.ok_or_else(|| anyhow!("no file 'operator.yaml' in EKO archive found"))?;
+
+            // NOTE: Since v0.15, EKOs are always in the flavour basis
```
cschwan, line 366: Move the directly above the initialization of the `pid_basis` member.

cschwan (changes requested):

pineappl_cli/src/evolve.rs
Radonirinaunimi: I also do not know the exact ranges so I simply added a comment as you suggested.

Radonirinaunimi:
This is now done! The test is failing because the data cache have not been updated. I would prefer to do so after https://github.com/NNPDF/pineappl/pull/344 is merged because that PR also unifies the caching action in **one** place, otherwise there will be some conflicts. So if you are fine with that PR, I will merge that first.

PS: I have update the Python script in master.

pineappl_cli/src/evolve.rs
```diff
@@ -70,6 +70,7 @@ mod eko {
         V0(MetadataV0),
         V1(MetadataV1),
         V2(MetadataV2),
+        V3(MetadataV3), // v0.15 - v????
```
felixhekhorn, line 73: actually: this would hopefully correspond to `__data_version__ = 3`, rather then the API version `v0.15`, so in the end I think it is a good name :innocent:

PR #353, opened 2025-06-24
Add `Grid::merge_channel_factors` and `Channel::factor`

cschwan:
**TODO:**
- [x] Add a dedicated FK table to use as a test
- [x] fix test coverage


cschwan:
Note that `FkTable::grid_mut` should not exist, because it allows you to break the invariants that make a grid an FK-table. Instead we should `.into_grid`, perform the operations and then finally convert back.

cschwan:
Maybe not. Try something like

    self = Self::try_from(self.into_grid().op).unwrap();

Radonirinaunimi:
> Maybe not. Try something like
> 
> ```
> self = Self::try_from(self.into_grid().op).unwrap();
> ```

This also doesn't work. Here, calling `into_grid()` is the problem as it seeks to move object from a mutable reference.

I will push first the changes with the methods also appearing in the FK table objects and then will think more about a better solution as soon as there is more time.

Radonirinaunimi:
I've spent some time again trying to find a better alternative but without much success. I think this is the best we can do.

cschwan:
I had some time to think about this and I believe the problem is simply that `FkTable::rotate_pid_basis` breaks our invariants - it operates on an `FkTable`, but the result is no longer an `FkTable` but a more general `Grid`. That was also your initial problem report!

We can fix that by wrapping `Grid::rotate_pid_basis` within `FkTable::rotate_pid_basis`, in which the latter
1. calls the former and then immediately
2. `split_channels`, then
3. `merge_channel_factors` and finally
4. `optimize_using(GridOptFlags::MERGE_SAME_CHANNELS)`.

All additional methods in `FkTable` should be removed, because they're either unnecessary (`split_channels` and `merge_channel_factors` - `FkTables` have channels with a single entry and the factor `1.0`, so both methods won't do anything) or allow to break invariants (`grid_mut`, ...).

Radonirinaunimi:
Thanks @cschwan, these make perfect sense! I tested this and it works (in the sense that one can read the dumped rotated FK table) only when using `optimize_using(GridOptFlags::all())`.

pineappl_py/tests/test_grid.py
```diff
@@ -399,6 +399,36 @@ def test_toy_convolution(self, fake_grids, params, expected):
...
+
+        # merge the factors and check that the channels are to unity
+        g.split_channels()
+        g.merge_channel_factors()
+        # optimize the grid to remove duplicate channels
+        g_facs = g.channels_factors()
+        np.testing.assert_array_equal(g_facs, 1)
+
+        # check that the FK table can be loaded properly
+        with tempfile.TemporaryDirectory() as tmpdir:
```
felixhekhorn, line 428: just to say [the recommended way of dealing with temporary files in pytest](https://docs.pytest.org/en/stable/how-to/tmp_path.html) is using the `tmp_path` fixture

pineappl_py/tests/test_boc.py
```diff
@@ -59,9 +59,10 @@ def _generated_bwfl_fields(n_bins: int, n_dimensions: int) -> BwflFields:
 
 class TestChannel:
     def test_init(self):
-        le = Channel([([2, 2], 0.5)])
-        assert isinstance(le, Channel)
-        assert le.into_array() == [([2, 2], 0.5)]
+        channels = Channel([([2, 2], 0.5)])
```
felixhekhorn, line 62: - could you add a second more complicated case with two entries? I'm not entirely sure what the function does ... 
- can you also clarify the name: should the variable be singular `channel` ? as in uubar channel here. I.e. the functions acts only _within_ a given channel, such as u-like-ubar-like (and not across to gluon-gluon)

pineappl/src/grid.rs
```diff
@@ -1482,6 +1483,21 @@ impl Grid {
             })
             .collect();
     }
+
+    /// Merges the factors of the channels into the subgrids to normalize channel coefficients.
+    ///
+    /// This method factors out the smallest absolute coefficient from each channel using
+    /// [`boc::Channel::factor`] and then scales the corresponding subgrids by these factors.
+    pub fn merge_channel_factors(&mut self) {
```
felixhekhorn, line 1491: Adding a unit test for this should be trivial as the action is supposed to be an identity (on the convolution numbers), right?

pineappl_py/tests/test_boc.py
Radonirinaunimi: I made it a bit clearer in [dc22670](https://github.com/NNPDF/pineappl/pull/353/commits/dc22670cd3355211489b38baed6a6d15b0197659). Hope it is better now.

Here, the `Channel` object can contain multiple channel combinations. For instance, for a $q \bar{q}$-initiated channel, it'll contain all the $n_f$ active combinations. So in this sense it can be singular or plural.

pineappl/src/grid.rs
Radonirinaunimi: What exactly do you mean here (w.r.t. the identity)?

felixhekhorn: take a (already existing) unit test for calling `convolute_with_one`:
```rust
pub fn test_convolute_with_one() {
 // create grid ...
 // call convolute_with_one ...
 // check numbers ...
 // ADD HERE: call merge_channel_factors
 // ADD HERE: check the numbers are still the same - i.e. identical to above
}
```
something like this should work, no?

Radonirinaunimi: Oh, yes, ofc. This operation is invariant under the convolution. We no longer have unit tests with `convolute_with_one` and this can only really be checked at the level of the Python API, so I added some in [f07d9f0](https://github.com/NNPDF/pineappl/pull/353/commits/f07d9f03f266f721d5a6bdaf125985027f886d61).

pineappl_py/tests/test_grid.py
```diff
@@ -399,6 +398,51 @@ def test_toy_convolution(self, fake_grids, params, expected):
...
+        g.split_channels()
+        g.merge_channel_factors()
+        # optimize the grid to remove duplicate channels
+        g_facs = g.channels_factors()
+        np.testing.assert_array_equal(g_facs, 1)
+
+        # check that the FK table can be loaded properly
+        path = f"{tmp_path}/grid_merged_factors.pineappl.lz4"
+        g.write_lz4(path)
+        g_mod = Grid.read(path)
```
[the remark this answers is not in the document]
Radonirinaunimi: Sorry, my bad 🙈 😅 I was in a rush to the point that it swept my head that everything is being modified in place. Thanks a tone for checking carefully!

I have fixed this by using the pre-computed numbers used in other tests so that we don't have to perform the convolution operation twice.

pineappl_py/src/boc.rs
```diff
@@ -274,6 +274,17 @@ impl PyChannel {
...
+    ///
+    /// Returns
+    /// -------
+    /// list(float) :
+    ///     list containing the factor values
+    #[must_use]
+    pub fn factors(&self) -> Vec<f64> {
+        self.entry.entry().iter().map(|(_, f)| *f).collect()
+    }
+
```
cschwan, line 287: Is this (convenience) method really needed? I'm always in favour of keeping APIs as small as possible (but you may disagree); one can also use `.into_array()` and then extract the factor from the Python side. AFAIU you use this method to check that the FK-table factors are correct, but this is now guaranteed.

pineappl_py/src/fk_table.rs
```diff
@@ -175,6 +175,22 @@ impl PyFkTable {
...
+    #[must_use]
+    pub fn channels_factors(&self) -> Vec<f64> {
+        self.fk_table
+            .grid()
+            .channels()
+            .iter()
+            .flat_map(|entry| entry.entry().iter().map(|(_, f)| *f))
+            .collect()
+    }
+
```
cschwan, line 193: Same remarks applies here.

pineappl_py/src/grid.rs
```diff
@@ -692,6 +692,21 @@ impl PyGrid {
...
+    ///     list containing the factor values
+    #[must_use]
+    pub fn channels_factors(&self) -> Vec<f64> {
+        self.grid
+            .channels()
+            .iter()
+            .flat_map(|entry| entry.entry().iter().map(|(_, f)| *f))
+            .collect()
+    }
+
```
cschwan, line 709: Same remarks applies here.

pineappl_py/src/boc.rs
cschwan: Yes, I suggest we don't add anything we don't need (yet), especially if the functionally is already there. If you remove them, we should be able to merge this quickly.

PR #354, opened 2025-06-26
Bugfix and refactor `Grid::evolve`

cschwan:


Radonirinaunimi:
How much of a bug was this actually?

PR #359, opened 2025-07-24
Add `set_subgrid` interface to the C-API

Radonirinaunimi:
Adds to the C-API the support to fill grids by setting directly the subgrids. This is needed for (SI)DIS-like processes where coefficient functions are computed.

Hi @vbertone, here is the interface with an example. Please let me know if this (does not) work(s).


Radonirinaunimi:
I will now merge this and make a `v1.3.1-alpha1` release in order for https://github.com/QCDLab/PineAPFEL to be able to use it in the CI. This is pretty much well-tested there.

Issue #338, opened 2025-04-10
Zero-scaling and subsequently optimizing any non-empty grid leads to a panic


cschwan:
This was found by @scarlehoff, who used

    pineappl write ${in_grid} ${out_grid} --delete-bins 1-${bins} --scale 0.0 --set-bins ${left},${right}

to create a new and empty grid with a single bin with limits `${left}` and `${right}`. This works, but optimizing the grid leads to a panic: <https://pastebin.com/AZtFbptU>.

scarlehoff:
fwiw, also reading it with the python API and calling `.table()` raises the error.

cschwan:
The panic comes from the fact that when optimizing a `Subgrid` we calculate the smallest range for each interpolation dimension that still captures all non-zero values. If the `Subgrid`, however, is completely empty, the constructed range is of the form `n..0` with `n` a positive non-zero number. This range is empty, but when write `values[range.clone()]` in line 216,

https://github.com/NNPDF/pineappl/blob/ee5da9b1cdbaa6ab2a232e675933bb3c083fcf3e/pineappl/src/subgrid.rs#L212-L217

this doesn't return an empty slice but produces the reported panic.

cschwan:
The actual problem is that the invariant

    array.is_empty() == array.indexed_iter().next().is_none()

for objects `array` of type `PackedArray` was previously violated when scaling with zero. In commit 591cdcfa434ef7028dbbdc5f8da2ab83b273029c we simply capture this special case and clear the array.

cschwan:
@scarlehoff: Could you please test `master` with the latest commit? It should definitely fix the panic from `Grid::optimize` and if we're lucky also the one from `FkTable::table`.

cschwan:
@scarlehoff: Note that for testing you need to start from non-empty grids, if you keep using the zero-scaled grids you should run into the same problems!

scarlehoff:
I've tested with [this script](https://pastebin.com/fTqzCkxm) and at least --optimize seems to work ok now, thanks :)

>  if we're lucky also the one from FkTable::table.

I am afraid we are not lucky :(

Note that this issue also occurs for old FkTables (which I believe were ported from old applgrids). I've uploaded one of the grids from th 700 to the server:

```bash
[...]

PR #360, opened 2025-08-08
Fix old zero scaled grids

felixhekhorn:
https://github.com/NNPDF/pineappl/issues/338 revealed a bug when multiplying a grid with 0 (i.e. default value), which got fixed by https://github.com/NNPDF/pineappl/commit/591cdcfa434ef7028dbbdc5f8da2ab83b273029c . However, this does not fix grids, which got multiplied before before that commit - which still yields to problems, see https://github.com/NNPDF/pineko/issues/228 .

The proposed solution is explained [here](https://github.com/NNPDF/pineko/issues/228#issuecomment-3136895074)

(I don't think I can do this on my own - I'm not an experienced enough PineAPPL developer myself, but I want to give an incentive for others to jump in :innocent: )

I think iterating on all subgrids and call `packed_array::clear_if_empty` might be enough

### DIFF
--- a/pineappl/src/packed_array.rs
+++ b/pineappl/src/packed_array.rs
@@ -1014,20 +1014,19 @@ mod tests {
     fn clear_if_empty() {
         let mut array = PackedArray::new(vec![40, 50, 50]);
 
+        // set something, which is not nothing
         array[[0,0,0]] = 1;
-
+        assert!(!array.is_empty());
+        array.clear_if_empty();
         assert!(!array.is_empty());
 
         // setting the default value does not clear the array on it's own ...
-
         array[[0,0,0]] = 0;
-
         assert!(!array.is_empty());
         assert_eq!(array.indexed_iter().count(),0);
 
         // ... one needs to make that explicitly
         array.clear_if_empty();
-
         assert!(array.is_empty());
         assert_eq!(array.indexed_iter().count(),0);
     }
--- a/pineappl/src/packed_array.rs
+++ b/pineappl/src/packed_array.rs
@@ -90,6 +90,18 @@ impl<T: Copy + Default + PartialEq> PackedArray<T> {
             .map(|(indices, entry)| (indices, *entry))
     }
 
+    /// Clear array, if it is empty.
+    /// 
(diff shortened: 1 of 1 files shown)


felixhekhorn:
Questions:
1. shouldn't `--repair` be an option of `pineappl write`? i.e. if I follow "never touch the input" I need a target and `write` is the place to be
2. should we use the already present and empty `--upgrade`?

cschwan:
> Questions:
> 
>     1. shouldn't `--repair` be an option of `pineappl write`? i.e. if I follow "never touch the input" I need a target and `write` is the place to be

Yes, exactly.

> 
>     2. should we use the already present and empty `--upgrade`?

No, `--upgrade` upgrades the internal data structures to newest available versions.

cschwan:
Thank you @felixhekhorn, for starting this! I added the functionality in commit 5533f4cda552b44ac4c3f2d4ac3ca3994c3cc923 and the CLI switch in commit 829626de7e0d97a2cd0f03eb11c7b2bcaf6c04d9.

To finish the PR, do we need a Python wrapper to `Grid::repair`? I suppose we need to answer whether we want to unconditionally repair grids in Pineko (potentially expensive), then we need the wrapper, or we want to repair the grids on a case-by-case basis. The latter has the advantage that we always know when something is going wrong.

felixhekhorn:
I see you took a quite different approach @cschwan then myself, which is proof again how different people approach problems. I consider that a strength of development, because e.g. in this case your version is much shorter :+1: 

as for the API: I'm tempted to say, that repair should be an intrinsic part of `optimize`, because using the current file version _is_ an optimization, i.e. `Grid::optimize` should call `Grid::repair` (potentially upon a optimization flag, which defaults to true). I think assuming that the current file version is better then the old is a safe one. The (potential) computational cost of `repair` and `optimize` are comparable, so we can join them.

Radonirinaunimi:
I would also prefer `repair` to be a separate method. And in that regard, I guess that would be the only piece missing before we can merge this, at least in order to propagate into `pineko` so that people can compute EKOs again.

felixhekhorn:
No idea on what the problem is, you did merge master but he still refuses to "rebase and merge" and indeed on the console I had to fix a merge conflict. However, then I could merge this branch cleanly on top of master - see there. Still Github didn't realize this here, so I close this PR any ways ...

PR #365, opened 2025-11-18
Fix panic in `fx2`

peterkrack:
I had some issue when calling pineko theory opcards for two of my grids, when the value y = 6.786550974400577 was passed. The newton method here would not converge in 10 steps.

Now using https://crates.io/crates/lambert_w

### DIFF
--- a/examples/fortran/lhapdf_example.output
+++ b/examples/fortran/lhapdf_example.output
@@ -1,6 +1,6 @@
  Computing predictions with the same PDF: NNPDF31_nlo_as_0118_luxqed
-   7.1330573367785985E-005   2.1399172010335785E-004
+   7.1330573367785999E-005   2.1399172010335790E-004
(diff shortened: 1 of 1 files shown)


peterkrack:
Changing all the check to be compliant with the solution using the lambert W function is pretty hard. Also the functions converning the compatibility with appl grid files seem to get compromised.

Anyway the issue is related to the FMA Fused-multiply-add-operation. Compiling that function in C or Fortran works only on x86. On arm/M1pro the function will fail for certain values and oscillate if using -ffast-math with -ffp-contract=on. On arm/M1 -ffast-math -ffp-contrac=off the function works. The same oscillating behaviour happens in Rust, which does (and should in general imho) do the FMA optimisation. Maybe there is a way to have just this function unoptimised?

[main.c](https://github.com/user-attachments/files/23614374/main.c)
[mainf90.txt](https://github.com/user-attachments/files/23614341/mainf90.txt)

Issue #346, opened 2025-05-12
Partially close convolutions in v1


felixhekhorn:
Is it possible in v1 to turn SIDIS grids to DIS grids XOR FF grids? i.e. can I close some convolution index, but still keep others open and dump this back into a grid file?

(Maybe this is said already somewhere?)

cschwan:
You cannot change the number of convolutions of a grid, but you can change [the convolution types](https://docs.rs/pineappl/1.0.0-alpha4/pineappl/grid/struct.Grid.html#method.convolutions_mut). What's your use case?

felixhekhorn:
> You cannot change the number of convolutions of a grid, but you can change [the convolution types](https://docs.rs/pineappl/1.0.0-alpha4/pineappl/grid/struct.Grid.html#method.convolutions_mut).

Mmm that's not very useful I think ... unless it was mislabelled there is no point in changing as different types represent different physics

> What's your use case?

turn a SIDIS grid into a DIS, specifically find an argument to convince [these people](https://inspirehep.net/literature/2788804) to add a PineAPPL interface ... and maybe investigate the impact in a PDF fit

I can see this might be technically difficult and the way out would be to have a fixed distribution laying around that you have to call the moment you're eventually can close all indices

felixhekhorn:
That is precisely what I mean. Just to be clear: `evolve` _already_ returns a grid as FK tables are special grids

cschwan:
Yes, but FK-tables are even more special than the grids we'd return and FK-tables don't change the number of convolutions (which I forgot previously). We probably need a new operation, which we could call `partial_convolve`. This operation should

- implement the partial convolution with exactly *one* convolution function (that's the most important case and the general case can be constructed from it) and
- return a new `Grid`, with one less interpolation, one less x and possibly one less Q2 parameter and so on.
- If we start with a grid that has just a single convilution, this operation must signal an error.

@t7phy: I think we talked about exactly this when we implemented the new file format, do remember anything from our discussion?

t7phy:
@cschwan tbh I don't remember a discussion for this particular situation. the most similar discussion I can remember had to do with convolution with EKOs, wherein if for example, we have a SIDIS grid, then do we:
- first convolve the grid with space-like eko and produce an intermediate  half-grid-half-fktable and then further convolve this with time-like eko? or
- do we pass both the ekos at one go to go directly from the Grid to the Fktable?

cschwan:
[...]

PR #368, opened 2026-01-11
Implement a `fix_convolution` method

Radonirinaunimi:
The following PR allows fixing one of the convolutions and return a new Grid with lower dimension. For eg. it can be used to convert a SIDIS grid into DIS.

**TODO:**
- [x] simplify a bit the logic
- [x] improve the error handling and CLI interface
- [x] perform extensive checks (checked in DY-like and single-inclusive hadron production cases)
- [x] propagate into the APIs

cc @vbertone

### DIFF
--- a/pineappl_py/tests/test_grid.py
+++ b/pineappl_py/tests/test_grid.py
@@ -600,6 +600,80 @@ class TestGrid:
             expected_results,
         )
 
+    def test_fix_convolution_dis(self, fake_grids, pdf):
+        channels = [Channel([([i], 0.1)]) for i in range(-5, 5)]
+
+        g = fake_grids.grid_with_generic_convolution(
+            nb_convolutions=1,
+            channels=channels,
+            orders=ORDERS,
+            convolutions=[CONVOBJECT],
+        )
+
+        with pytest.raises(BaseException) as err_func:
+            g.fix_convolution(conv_idx=0, xfx=pdf.unpolarized_pdf)
+        assert "cannot fix the last convolution" in str(err_func.value)
+
+    def test_fix_convolution_wrong_index(self, fake_grids, pdf):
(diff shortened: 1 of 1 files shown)


pineappl_cli/src/write.rs
```diff
@@ -211,6 +214,28 @@ impl FromArgMatches for MoreArgs {
...
+                                            ),
+                                        )
+                                    })?;
+                                if !(1..=3).contains(&conv_idx) {
+                                    return Err(Error::raw(
+                                        clap::error::ErrorKind::InvalidValue,
+                                        "Convolution index must be 1, 2, or 3",
+                                    ));
+                                }
+                                OpsArg::FixConvolution((conv_idx - 1, arg[1].clone()))
```
felixhekhorn: Does it make sense / is it consistent to be 1-based?

Radonirinaunimi: This, I don't really have a strong preference. On the other hand, using a zero-based convention for a command line argument is a bit weird. Also, I don't think it is best to prepend with `xN` as suggested in #346.

pineappl/src/grid.rs
```diff
@@ -1437,6 +1438,166 @@ impl Grid {
...
+            (self.orders.len(), self.bwfl.len(), new_channels.len()),
+            || EmptySubgridV1.into(),
+        );
+
+        let conv_to_fix = &self.convolutions[conv_idx];
+
+        for (inew_chan, new_pids) in new_channel_pids.iter().enumerate() {
+            let origins = &new_channel_map[new_pids];
+
+            for (iord, _) in self.orders().iter().enumerate() {
```
cschwan: `for iord in 0..self.orders().len()` is more idiomatic.

cschwan (changes requested):

PR #374, opened 2026-02-23
Fix failing release action

### LINKED ISSUE #373: Dropping `PyPy3.7` from `wheels-{macos,windows}`
The CI release has failed because the previously used macos runners have reached EOL. I updated them to use `macos-15` but then PyPy 3.7 is no longer available in aarch64 macos.

I believe there isn't other way than dropping this?

cschwan:
Implements changes to close #373.


Radonirinaunimi (approved):
These are indeed the changes I had in mind.

PR #379, opened 2026-03-05
Automatically install the OOP C++ header

Radonirinaunimi:
So far, the OOP C++ header had to be copied manually into where the C header is located or within the project. This allows one to specify a `PINEAPPL_C_INSTALL_PREFIX` environment variable so that the header is automatically copied at build time. The release workflow has also been adjusted so that this is automatically installed when copying the pre-built binaries:

```sh
curl --proto ... | sh -s -- --prefix /my/custom/installation/path
```

 **TO DO:**
- [x] Update the documentation

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -75,6 +75,7 @@ If you want to build the CAPI from its sources instead, you first need to
 3. Now install `pineappl_capi`, PineAPPL's C API:
 
        cd pineappl_capi
+       export PINEAPPL_C_INSTALL_PREFIX=${prefix} # To copy the OOP C++ header
        cargo cinstall --release --prefix=${prefix} --libdir=${prefix}/lib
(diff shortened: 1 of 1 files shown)


Radonirinaunimi:
>     * out of laziness I haven't explicitly run the code to check if it actually works - should I do so?
>
No, not needed. Everything is being checked in the CI. So it was more whether or not you agree with the concept.
 
>     * I would be happy if the OOP interface gets integrated more into the program, but if I remember correctly from [Stabilize APIs #328](https://github.com/NNPDF/pineappl/pull/328) @cschwan is sceptical about that ... in particular, it was a conscious decision to leave the OOP interface aside; also due to compatibility reasons

I believe (if I recall correctly) that this discussion was RE the Fortran side.

Issue #372, opened 2026-02-05
Panic! At the optimization (for pineappl > 1)


scarlehoff:
I'm getting a panic at optimization time (I believe, looking at the exception, see below).
This only happens for versions of pineappl above > 1

I have uploaded a grid that fails to optimize (even from the CLI), just rename `.zip` to `.lz4`

I've tested the following combinations:

```
pineappl-0.8.7-py313h5e81bce_4.conda - Works
pineappl-1.3.0-py313h7ef63f5_0.conda - Does not work
pineappl 1.3.0 # installed with the script, does not work
```

The stack backtrace below is for 1.3.0
```
thread '<unnamed>' panicked at pineappl/src/interpolation.rs:34:9:
internal error: entered unreachable code
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at core/src/panicking.rs:221:5:
panic in a function that cannot unwind
stack backtrace:
   0:     0x7fcdcf554bf9 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hc002032c745ec399
   1:     0x7fcdcf4c260b - core::fmt::write::h78425ec9962d4be7
   2:     0x7fcdcf525cbe - std::io::Write::write_fmt::ha00b146e89c56092
   3:     0x7fcdcf555ff2 - std::sys::backtrace::BacktraceLock::print::h59ceb186f016850e
   4:     0x7fcdcf555e5b - std::panicking::default_hook::{{closure}}::h9fcfa010b5ecd19c
   5:     0x7fcdcf5565b6 - std::panicking::rust_panic_with_hook::h0a4b940661a26423
   6:     0x7fcdcf556122 - std::panicking::begin_panic_handler::{{closure}}::h72160a9fdca8fdc8
   7:     0x7fcdcf5560b9 - std::sys::backtrace::__rust_end_short_backtrace::hb6826ea75cd9c94b
   8:     0x7fcdcf5560a4 - rust_begin_unwind
   9:     0x7fcdcf4c59d4 - core::panicking::panic_nounwind_fmt::he104e461f7e92412
  10:     0x7fcdcf4c5a3c - core::panicking::panic_nounwind::hef5dfaaaad24ed17
  11:     0x7fcdcf4c59e4 - core::panicking::panic_cannot_unwind::h44cd3738aca63805
  12:     0x7fcdcf4ab1f2 - pineappl_grid_optimize
[...]

PR #380, opened 2026-03-12
Various API fixes

Radonirinaunimi:
This PR addresses the various minor issues raised by Alex Huss. In particular, it:

- [x] provides an actual implementation of `pineappl_grid_metadata` and `pineappl_grid_set_metadata`
- [x] fixes `_channels_add` in the Fortran API
- [x] addresses https://github.com/NNPDF/pineappl/issues/372 (increase the max iteration from 10 to 15 to account for floating-point jitter)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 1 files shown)


Radonirinaunimi:
> This looks good, but it would be nice to have regression test checking Alex's and Juan's cases the require the fix. Would you please add them if it's not too complicated?

Yes, that's right! I just added both a unit test and grid to check this. Thanks.

PR #398, opened 2026-04-15
Fix pure Fragmentation scale evolution

Radonirinaunimi:
Sorry for having been slow lately which has resulted in https://github.com/NNPDF/pineappl/pull/394 being blocked. In the meantime, I stumbled upon an issue that has to do with evolving pure FF grids, ie with `fac = NoScale` and only fragmentation scale involved. This PR provides a fix for this and adds a test.


pineappl_py/src/evolution.rs
```diff
@@ -94,9 +94,16 @@ impl PyEvolveInfo {
     }
 
     /// Squared factorization scales of the `Grid`.
+    ///
+    /// Note: for pure time-like (FF) grids, PineAPPL stores only fragmentation scales. In that
+    /// case, this getter returns `frg1` as a backwards-compatible fallback.
     #[getter]
     fn fac1<'py>(&self, py: Python<'py>) -> Bound<'py, PyArray1<f64>> {
```
cschwan, line 101: This change will result in the Python and Rust APIs behaving differently. Is this justified?

cschwan (changes requested):
I believe the TODO comment,
```rust
// TODO: implement evolution for non-zero fragmentation scales
```
signifies that any evolution with fragmentation function hasn't been tested yet. Is that the case and is you testing SIA just the first instance of finding something wrong?

Radonirinaunimi:
> I believe the TODO comment,
> 
> ```rust
> // TODO: implement evolution for non-zero fragmentation scales
> ```
> 
> signifies that any evolution with fragmentation function hasn't been tested yet. Is that the case and is you testing SIA just the first instance of finding something wrong?

The comment indeed mentions that we FF scale evolution has not been tested at all. The reason was that until now we did not really grids and evolution operators to really test it. Now that we have PineAPFEL, we can easily test this for not only SIA but also for SIDIS, etc. The issue originally manifested from SIA but noticed it also for SIDIS. And this PR solves both.

pineappl_py/src/evolution.rs
Radonirinaunimi: Yes, I'd say that this is justified given how minimal this change is compared to say accounting it at the level of say `pineko` which would introduce non-negligible breaking changes.

Radonirinaunimi:
@cschwan Are you perhaps planning to look into this PR or should I merge it?

cschwan:
@Radonirinaunimi: merge it!

Issue #291, opened 2024-06-02
Is the pineappl cli printout a stable well defined output?


scarlehoff:
The `PineaAPPL` cli is much faster than e.g., the python interface. As a result, sometimes it is quite convenient to use it and then parse the results.

I'm currently parsing the cli instead of using `convolute_with_one` (with or without scales). This seems to work in most (all?) cases. Can I trust this output to be constant in time? If now, can we have some kind of `--stable-output` that will be stable in time?

At the moment my assumptions for `uncert` and `convolve` are:
1. The three first lines are part of the header.
2. The results are always the last columns and there's nothing beyond.

For reference, this is the code I'm using:
```python
# before 0.74 you need to add `--silence-lhapdf`
def _convolute_with_pineappl_cli(grid_path, pdf_name, member):
    """Use the pineappl cli to compute the convolution of the grid with the given pdf

    Equivalent to:
        pineappl convolve <grid_path> <pdf>
    """
    cmd = ["pineappl", "convolve", grid_path, f"{pdf_name}/{member}"]
    result_raw = sp.run(cmd, stdout=sp.PIPE)
    results = result_raw.stdout.decode("utf-8").strip()
    predictions = [line.split()[-1] for line in results.split("\n")[3:]]
    return np.array(predictions, dtype=float)


def _scales_with_pineappl_cli(grid_path, pdf_name, n_scales):
    """Use the pineappl cli to compute the convolution of the grid with the given pdf

    Equivalent to:
        pineappl uncert <grid_path> <pdf> --scale-abs=9
    """
    cmd = ["pineappl", "uncert", grid_path, pdf_name, f"--scale-abs={n_scales}"]
    result_raw = sp.run(cmd, stdout=sp.PIPE)
    results = result_raw.stdout.decode("utf-8").strip()
    predictions = [line.split()[-n_scales:] for line in results.split("\n")[3:]]
    return np.array(predictions, dtype=float)
```

cschwan:
It's not guaranteed to be stable, but so far the changes have been minimal between versions.

I'm thinking of making the output stable at some point and furthermore prefixing each line that's not data with `#` so you can simply parse the output as an ascii table.

PR #400, opened 2026-05-05
Add `--discard-non-matching-values` to remove non-matching `x` and scales when exporting into APPLgrid

Radonirinaunimi:
TODO:
 - [x] Add a CI test

### DIFF
--- a/.github/actions/cache-test-data/action.yml
+++ b/.github/actions/cache-test-data/action.yml
@@ -10,7 +10,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: test-data
-        key: test-data-v31
+        key: test-data-v32
     - name: Download test data if cache miss
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
--- a/.github/actions/cache-test-data/action.yml
+++ b/.github/actions/cache-test-data/action.yml
@@ -10,7 +10,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: test-data
-        key: test-data-v30
+        key: test-data-v31
     - name: Download test data if cache miss
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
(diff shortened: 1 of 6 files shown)


cschwan (commented):
Looks good to me, although we should refactor the code at some point, but probably not here and now.

I wonder whether we should merge the option into a single `discard-non-matching-values` that enables both. I don't think there's a reason to keep them separate. To make it backwards-compatible, we should add an [alias](https://docs.rs/clap/latest/clap/struct.Arg.html#method.alias) (see also #291).

Radonirinaunimi:
I fully agree! And I just implemented it in [8ba5416](https://github.com/NNPDF/pineappl/pull/400/commits/8ba54161e590dc7289a92e321b48bb748d0376e0).

cschwan (changes requested):
Can you extract the problematic subgrid(s)? The remaining subgrids are just dead weight.

Radonirinaunimi:
I am not exactly sure what you mean by this?

cschwan:
In general, we should keep the test-data size as small as possible, also to minimize the test run time. In this case I think we can just isolate the subgrid that's affected by the new flag and throw away the rest.

cschwan:
Commit f928b80ac2a2d1ee1a92ca132b2235ec8a3f6bf0 uses a smaller grid, which I created by stripping all log-orders and all bins except bin # 63. This bin has the most 'failing' values, including both factorization-scale and momentum-fraction types; the test grids goes down in size from 175 MB to 216 KB, which saves bandwidth and testing/CI time.

cschwan (approved):
@Radonirinaunimi: have a look at my changes; as far as I am concerned this can be merged.

Radonirinaunimi:
Sorry @cschwan I was not able to look into it yesterday, so thanks for taking care of it. The changes look perfect to me.

PR #406, opened 2026-06-19
Revise citation section in README.md

t7phy:
Updated citation section to include references for both version 0 and version 1 of the software.


cschwan (approved):
LGTM!

